PR #1, opened 2016-03-29
(SERVER-1161) Add travis

rlinehan:


rlinehan:
I think this should be ready to go now.

fpringvaldsen:
:see_no_evil:

project.clj
```diff
@@ -0,0 +1,19 @@
...
+
+  :lein-release {:scm         :git
+                 :deploy-via  :lein-deploy}
+
+  :deploy-repositories [["releases" {:url "https://clojars.org/repo"
+                                     :username :env/clojars_jenkins_username
+                                     :password :env/clojars_jenkins_password
+                                     :sign-releases false}]]
+
+  :plugins [[lein-release "1.0.5"]])
```
cprice404: I don't think this is necessary, lein now has the release task built-in.  We probably haven't cleaned this up from other projects where we added it before that version of lein.

cprice404:
+1 other than removing that release plugin.

project.clj
rlinehan: Aha! That's good to know. What about the stuff on line 11 (the `:lein-release` key)? Seems like that would still be necessary to tell lein-release what to do?

rlinehan: Although actually, looking at the documentation at https://github.com/technomancy/leiningen/blob/master/doc/DEPLOY.md this might not be needed after all.

rlinehan:
@cprice404 Updated!

project.clj
cprice404: Hmm, good catch on the `:lein-release` stuff, I don't think it'd ever occurred to me that that might not be needed.

PR #2, opened 2016-04-12
(SERVER-1162) initial jruby code

cprice404:
This PR sets up the initial skeleton of the jruby code for this repo.

Prior to filing the PR, I basically forcefully reset the history of this repo to that of the puppet-server master branch.  The goal was to retain the git history for the jruby-related files that we are keeping here.

The first commit here simply deletes the vast majority of the files from the puppetserver repo, leaving behind only the ones that we will want to keep.  After that commit, the tests are totally broken and the jruby namespaces won't compile because I did not make any changes to their contents, in hopes of making this PR slightly more feasible to review.

The second commit brings back all of the docs, travis scripts, and basic changes to the project.clj file that @rlinehan had originally set up in this repo.  It contains modifications to the project.clj file that are necessary to get the tests running (though they will still fail after this commit).

The third commit is the meat of this PR.  It refactors all of the jruby-related namespaces and tests to get them passing.  The vast majority of this commit is the removal of Puppet-specific constructs such as things relating to Puppet environments, settings, profiler, and HTTP client.  After this commit the tests pass, though there are still several more things we'll want to do in follow-up PRs; here's a possible list:

1. Change names of things to no longer reference "Puppet"
2. Get rid of references to TK?  Leaning towards making this more of a pure library, rather than providing a TK service.  I think that we might still keep the *protocol* around, but rather than having a TK service we might just have a `create-jruby-pool` function that would return an object that satisfied the protocol.  Need to poke at this a bit and validate that this strategy is feasible and that we're not losing anything very useful if we get rid of the TK service.
3. Add callback hooks for cleanup (and maybe initialization) so that downstream projects like puppetserver can add custom logic.
4. Fix up a few tests to add coverage that I've commented out for now, where the test was relying on the presence of the Puppet Ruby code even though it was testing something unrelated.
5. Consider moving the CLI wrappers (`gem.clj`, etc.) back into this repo, assuming that we can find some way to make them accessible to ezbake in downstream projects?


cprice404:
@rlinehan IMO this can be merged as an initial stab at this stuff even though there will be follow-up work to do.  I think we probably want @camlow325 to review it before it gets merged, so it's fine for you to prioritize it below your http-client stuff.  At some point if you get a chance, though, it'd be nice if you could just take a super quick peek at how I've structured the commits in general and let me know if you think it's tolerable in terms of reviewability and maintaining the history that you were interested in.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -174,56 +147,18 @@
...
-                     default-http-socket-timeout))
       (update-in [:compile-mode] #(keyword (or % default-jruby-compile-mode)))
       (update-in [:borrow-timeout] #(or % default-borrow-timeout))
-      (update-in [:master-conf-dir] #(or % default-master-conf-dir))
-      (update-in [:master-var-dir] #(or % default-master-var-dir))
-      (update-in [:master-code-dir] #(or % default-master-code-dir))
-      (update-in [:master-run-dir] #(or % default-master-run-dir))
-      (update-in [:master-log-dir] #(or % default-master-log-dir))
       (update-in [:max-active-instances] #(or % (default-pool-size (ks/num-cpus))))
-      (update-in [:max-requests-per-instance] #(or % 0))
```
rlinehan: Is `max-requests-per-instance` something that we want to keep as a jruby setting?

rlinehan: Oh wait, nevermind, it's still there, the diff's just a bit hard to read.

rlinehan:
@cprice404 I skimmed this briefly - this seems reasonable - thank you for doing the work to preserve the git history! I would definitely be in favor of merging this sooner and filing follow-up PRs, because the diff for this is already pretty hard to follow. So I'm 👍 .

project.clj
```diff
@@ -1,43 +1,27 @@
...
-                 [prismatic/schema "1.0.4"]
-                 [me.raynes/fs "1.4.6"]
-                 [liberator "0.12.0"]
-                 [org.apache.commons/commons-exec "1.3"]
-
                  [org.jruby/jruby-core "1.7.20.1"
                   :exclusions [com.github.jnr/jffi com.github.jnr/jnr-x86asm]]
                  ;; jffi and jnr-x86asm are explicit dependencies because,
-                 ;; in JRuby's poms, they are defined using version ranges,
+                 ;; icn JRuby's poms, they are defined using version ranges,
```
camlow325: s/icn/in?

project.clj
```diff
@@ -47,108 +31,27 @@
...
-  :source-paths ["src/clj"]
-  :java-source-paths ["src/java"]
-  :test-paths ["test/unit" "test/integration"]
-  :resource-paths ["resources" "src/ruby"]
-
-  :repositories [["releases" "http://nexus.delivery.puppetlabs.net/content/repositories/releases/"]
-                 ["snapshots" "http://nexus.delivery.puppetlabs.net/content/repositories/snapshots/"]]
-
-  :plugins [[lein-release "1.0.5" :exclusions [org.clojure/clojure]]]
+                 [org.clojure/tools.logging "0.3.1"]
```
camlow325: Just curious why you went with this version.  Looks like the version that the latest code in trapperkeeper and kitchensink use is version 0.2.6.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
```diff
@@ -5,13 +5,11 @@
             [puppetlabs.kitchensink.classpath :as ks-classpath]
```
camlow325: Nit - looks like this isn't used anymore.

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
```diff
@@ -149,25 +99,20 @@
...
+;; TODO: rename, get rid of references to puppet
 (schema/defrecord JRubyPuppetInstance
-                  [pool :- pool-queue-type
-                   id :- schema/Int
-                   max-requests :- schema/Int
-                   flush-instance-fn :- IFn
-                   state :- JRubyInstanceStateContainer
-                   jruby-puppet :- JRubyPuppet
-                   scripting-container :- ScriptingContainer
-                   environment-registry :- (schema/both
```
camlow325: Makes sense to remove this since it's Puppet-specific.  Seems like maybe we'll need a way - as part of the callback initialization hooks we'll do later - to stuff arbitrary custom state onto a JRubyInstance as needed?

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -227,10 +196,12 @@
...
 (deftest ^:integration hold-file-handle-on-instance-while-pool-flush-in-progress-test
   (testing "file handle opened from old pool instance is held open across pool flush"
-    (bootstrap/with-puppetserver-running
+    (tk-testutils/with-app-with-config
       app
-      {:jruby-puppet {:max-active-instances 2
-                      :borrow-timeout default-borrow-timeout}}
+      [jruby/jruby-puppet-pooled-service]
+      (jruby-testutils/jruby-puppet-tk-config
+       (jruby-testutils/jruby-puppet-config {:max-active-instances      4
```
camlow325: Does this need to be 4?  Seems like 2 was enough for this one but I can't remember very well anymore.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
```diff
@@ -27,15 +26,14 @@
   (jruby-testutils/jruby-puppet-tk-config
     (jruby-testutils/jruby-puppet-config {:max-active-instances pool-size})))
 
-(defn jruby-service-test-config-with-timeouts
+#_(defn jruby-service-test-config-with-timeouts
```
camlow325: Seems like we could just delete this.

camlow325:
Seems like the whole `jruby_request` namespace could be removed from the repo?

camlow325:
Don't have to do that on this PR but seems like we'll need to find some replacement for the `ruby-code-dir` that we hard-code in Puppet Server's `jruby-puppet-internal` namespace to `puppet-server-lib`.  Maybe that will need to be an external parameter in the new repo - or we force the client of the API in this repo to add that in to the `ruby-load-path` parameter so that the code in this repo doesn't have to have any special knowledge about it?

test/unit/puppetlabs/services/jruby/jruby_testutils.clj
```diff
@@ -20,12 +17,6 @@
 (def gem-home "./target/jruby-gem-home")
```
camlow325: Seems like we should be able to nuke the puppet-specific `ruby-load-path` from these tests.

camlow325:
:+1: I left some cosmetic feedback throughout.  Could just address that on subsequent cleanup PRs.  I'm good with just merging this as-is if you don't have any time to look into this further before you leave for PTO.

project.clj
cprice404: It was just the latest one on maven central and I wasn't aware of any reason not to use it.

src/clj/puppetlabs/services/jruby/jruby_puppet_core.clj
cprice404: Yeah that diff is super confusing... :(

src/clj/puppetlabs/services/jruby/jruby_puppet_schemas.clj
cprice404: Yeah, something like that... or like a callback that allows the user to modify the object after initialization, and return a modified copy that we will use in the pool.  In that case it could maybe be a subclass or something, and then they could decide how they want to handle the additional state on their own.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
cprice404: Dunno, but this whole test suite only takes like 30 seconds to run, so I'm not sweating it ATM.

test/unit/puppetlabs/services/jruby/jruby_puppet_service_test.clj
cprice404: Yep, good catch.

test/unit/puppetlabs/services/jruby/jruby_testutils.clj
cprice404: agree

cprice404:
> Seems like the whole jruby_request namespace could be removed from the repo?

Good call, deleting.

> we'll need to find some replacement for the ruby-code-dir that we hard-code in Puppet Server's jruby-puppet-internal namespace ... or we force the client of the API in this repo to add that in to the ruby-load-path parameter so that the code in this repo doesn't have to have any special knowledge about it?

Agree re: unnecessary special knowledge.  Deleting that stuff.

cprice404:
Added a commit that addresses @camlow325 's feedback.

camlow325:
Travis failed for JDK 7 only on this test with the latest changes...

```
lein test :only puppetlabs.services.jruby.jruby-pool-int-test/flush-jruby-pool-test

FAIL in (flush-jruby-pool-test) (jruby_pool_int_test.clj:165)

Flushing the pool results in all new JRuby instances
expected: (true? (verify-no-constants pool-context 4))
  actual: (not (true? false))
```

It seems like we've seen this before (intermittently) in the original code from the Puppet Server so I don't think this is new with this PR.  For now, I'm going to rekick this job and see if it passes.

camlow325:
The tests continue to fail after several Travis re-runs, failing on different JDKs each time.  I've identified a couple of race conditions in the tests and plan to fix those in a PR for [TK-364](https://tickets.puppetlabs.com/browse/TK-364).  For now, though, I'm going to merge this PR.

rlinehan:
As a follow-up on this, I filed TK-365, TK-366, and TK-367 for the first 3 things in the todo list @cprice404 wrote in the PR message.

PR #3, opened 2016-04-19
(TK-364) Cleanup jruby-pool-int-test race conditions

camlow325:
This commit addresses a few race conditions in the jruby-pool-int-tests,
including:

1) Ensure that watch for pool flush isn't registered until after pool
   initialization is completed.  Previously, initialization complete
   could be misinterpreted as flush complete.

2) Hold borrow counts on pool instances constant in `wait-for-new-pool`
   so that an individual instance flush does not occur as a side effect
   of waiting for a full pool flush and `wait-for-new-pool` doesn't
   misinterpret that individual instance flush as the full pool flush.

3) Ensure that pool agents are done processing before the app is
   shutdown in `max-requests-flush-while-pool-flush-in-progress-test`,
   avoiding the appearance of an annoying error log message that could
   pop up previously.

### DIFF
--- a/test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
+++ b/test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
@@ -56,26 +56,13 @@
 (def script-to-check-if-constant-is-defined
   "! $instance_id.nil?")
 
-(defn await-jruby-agents
-  "Wait for pending actions on agents associated with the jruby pool
-  service to free up."
-  [{:keys [pool-agent flush-instance-agent]}]
-  ;; Dump a dummy action into each of the jruby pool agents and await that
-  ;; action's completion.  This assumes that any previously existing
-  ;; actions would have been flushed through prior to the new one being
-  ;; processed.
-  (jruby-agents/send-agent pool-agent #(constantly true))
-  (await pool-agent)
-  (jruby-agents/send-agent flush-instance-agent #(constantly true))
-  (await flush-instance-agent))
-
 (defn add-watch-for-flush-complete
-  [pool-context]
+  [{:keys [pool-agent]}]
   (let [flush-complete (promise)]
-    ;; Make sure none of the pool-related agents are processing anything
+    ;; Make sure the pool agent isn't processing anything
     ;; before registering a watcher for flush completion
-    (await-jruby-agents pool-context)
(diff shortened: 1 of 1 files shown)


camlow325:
This is intended to clean up sporadic test failures seen with #2 in Travis. /CC @rlinehan.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -55,9 +56,25 @@
...
 
+(defn await-jruby-agents
+  "Wait for pending actions on agents associated with the jruby pool
+  service to free up."
+  [{:keys [pool-agent flush-instance-agent]}]
+  ;; Dump a dummy action into each of the jruby pool agents and await that
+  ;; action's completion.  This assumes that any previously existing
+  ;; actions would have been flushed through prior to the new one being
+  ;; processed.
+  (jruby-agents/send-agent pool-agent #(constantly true))
```
rlinehan: My reading of the documentation for await makes it seem like these dummy actions wouldn't be needed to ensure that all previous actions have been flushed through

```
Blocks the current thread (indefinitely!) until all actions dispatched thus far,
from this thread or agent, to the agent(s) have occurred.
```
and also like maybe you could just do `(await pool-agent flush-instance-agent)`, but maybe I'm missing something?

rlinehan:
👍 had one comment that is mostly me trying to understand how `await`/this test works.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
camlow325: Yeah, re-reading that comment now, I think you're right.  I had originally read that as `await` only blocking until actions that had originated from the calling thread - as opposed to any thread - had completed.  But when I tried this out locally, I see that even the flush actions originating from a webserver-controlled thread are flushed through from an `await` call on the main test thread.  Given that, I took out the dummy actions in 9ca5cd1.  Since the tests only need to ensure that the pool agent is flushed before registering a wait for flush completion and for the flush instance agent to be flushed through before shutdown in the one test that needs it, I just removed this helper function and replaced with the appropriate in-line awaits.

rlinehan:
LGTM, gonna merge. 🚢 @cprice404 if you want to do a post-hoc review when you get back, that would be cool.

PR #4, opened 2016-04-22
(TK-364) Another jruby-pool-int-test race condition cleanup attempt

camlow325:
This commit makes another attempt at cleaning up some transient race
condition failures which appear to be due to some tests misinterpreting
that a pool flush has completed when it is still in progress and,
therefore, failing when attempting to assert that fresh instances have
been put back into the pool.

Previous code added a watch on the pool agent and interpreted the
receipt of a callback on the watch as an indication that a flush had
succeeded.  It seemed possible - although not locally reproducible -
that the callback could be invoked prematurely.  This commit changes the
code to perform an await on the pool agent after the point that the
flush is in progress, assuming that the await will block until after the
flush should have completed.

### DIFF
--- a/test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
+++ b/test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
@@ -13,9 +13,9 @@
 
 (def default-borrow-timeout 180000)
 
-(defn timed-deref
-  [ref]
-  (deref ref 240000 :timed-out))
+(defn timed-await
+  [agent]
+  (await-for 240000 agent))
 
 (defn get-stack-trace-for-thread-as-str
   [stack-trace-elements]
@@ -55,19 +55,6 @@
 
 (def script-to-check-if-constant-is-defined
   "! $instance_id.nil?")
-
-(defn add-watch-for-flush-complete
-  [{:keys [pool-agent]}]
-  (let [flush-complete (promise)]
-    ;; Make sure the pool agent isn't processing anything
-    ;; before registering a watcher for flush completion
-    (await pool-agent)
-    (add-watch pool-agent :flush-callback
-               (fn [k a old-state new-state]
-                 (when (= k :flush-callback)
-                   (remove-watch a :flush-callback)
-                   (deliver flush-complete true))))
-    flush-complete))
 
 (defn set-constants-and-verify
   [pool-context num-instances]
@@ -169,11 +156,10 @@
             pool-context (:pool-context context)]
         ;; set a ruby constant in each instance so that we can recognize them
         (is (true? (set-constants-and-verify pool-context 4)))
-        (let [flush-complete (add-watch-for-flush-complete pool-context)]
-          (jruby-protocol/flush-jruby-pool! jruby-service)
-          (is (true? (timed-deref flush-complete))
-              (str "timed out waiting for the flush to complete, stack:\n"
-                   (get-all-stack-traces-as-str))))
+        (jruby-protocol/flush-jruby-pool! jruby-service)
+        (is (true? (timed-await (:pool-agent pool-context)))
+            (str "timed out waiting for the flush to complete, stack:\n"
+                 (get-all-stack-traces-as-str)))
         ;; now the pool is flushed, so the constants should be cleared
         (is (true? (verify-no-constants pool-context 4)))))))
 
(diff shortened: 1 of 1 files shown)


camlow325:
The changes in this PR also allow the jruby-pool-int tests to follow the pattern used in the jruby-puppet-agents tests.  For example: https://github.com/puppetlabs/jruby-utils/blob/d056e1cf404e701e8c092a7c5bf1efa027898a22/test/unit/puppetlabs/services/jruby/jruby_puppet_agents_test.clj#L40.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -315,14 +300,16 @@
            (is (true? (set-constants-and-verify pool-context 3)))
 
            ;; The flush should still not have completed at this point because
-           ;; the last instance from the old pool is still being borrowed
-           (is (not (realized? flush-complete)))
+           ;; the last instance from the old pool is still being borrowed.
+           ;; Wait for some arbitrary (but short) period of time to see that
+           ;; that the agent is still busy - presumably handling the flush.
+           (is (false? (await-for 50 pool-agent)))
```
rlinehan: This seems maybe kind of racy? Although I guess if it passes locally there's probably not much chance the flush would be faster (and therefore fail) in Travis?

camlow325: Yeah, it is racy - and I think you're right that it probably would be more likely to fail locally (and, therefore, help identify a problem with the flush) than in Travis with the relatively short timeout.  I don't think it's any more racy than what it's replacing, though.  I could make the wait longer to make it more likely that we'd catch a failure, but I'm also sensitive to drawing out the run time on this test too much since it seems likely that this test suite is going to get slower over time anyway.

I don't have any good ideas on how to make this more airtight and not sure if it's really worth it to spend more time thinking about it at this point.  What do you think?

rlinehan: 👍 I think that everything you're saying makes sense, and yeah, probably not worth spending more time on it at the moment.

rlinehan:
I'm good on this and gonna merge. ping @cprice404 if you are interested in post-hoc review when you get back.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -190,18 +176,18 @@
...
+                        :hold-instance-while-pool-flush-in-progress-test)]
           ;; trigger a flush
           (jruby-protocol/flush-jruby-pool! jruby-service)
           ;; wait for the new pool to become available
           (is (true? (wait-for-new-pool jruby-service)))
           ;; return the instance
           (jruby-protocol/return-instance jruby-service instance :hold-instance-while-pool-flush-in-progress-test)
           ;; wait until the flush is complete
-          (is (true? (timed-deref flush-complete))
+          (is (true? (timed-await (:pool-agent pool-context)))
```
cprice404: Isn't there a different race here now?  If the flush completes before we hit the `timed-await`, then won't that cause a problem?

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
cprice404: Hmm.  I guess this is sort of the same question I asked above, and I guess I'm OK with leaving it as-is if you two are.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
camlow325: I don't think this will be a problem.  Seems like we just have this line here to ensure that the flush is done by the time we move on to the next line?  I think the `timed-await` function should still return `true` even if the flush completed before the Clojure `await` call injects its own action into the `:pool-agent`.  Did you have a specific problem in mind?

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
camlow325: I agree there's a risk this assertion may pass even if the flush completes sometime after this line is run and the subsequent `return-instance` call is made.  If you have an idea on how to make that better, I'm definitely open to doing that.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
cprice404: OK, read the docs for `await-for`, and now I think you're right.

PR #5, opened 2016-04-22
(TK-365) Remove references to 'puppet'

rlinehan:
This commit renames all the things to remove
references to 'puppet'.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_core.clj
@@ -165,14 +165,14 @@
 
 (schema/defn ^:always-validate
   free-instance-count
-  "Returns the number of JRuby instances available in the pool."
+  "Returns the number of JRubyInstances available in the pool."
   [pool :- jruby-schemas/pool-queue-type]
   {:post [(>= % 0)]}
   (.size pool))
 
 (schema/defn ^:always-validate
   instance-state :- jruby-schemas/JRubyInstanceState
-  "Get the state metadata for a JRuby instance."
+  "Get the state metadata for a JRubyInstance."
   [jruby-instance :- (schema/pred jruby-schemas/jruby-instance?)]
   @(:state jruby-instance))
 
--- a/src/clj/puppetlabs/services/jruby/jruby_core.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_core.clj
@@ -1,13 +1,13 @@
-(ns puppetlabs.services.jruby.jruby-puppet-core
+(ns puppetlabs.services.jruby.jruby-core
   (:require [clojure.tools.logging :as log]
             [schema.core :as schema]
             [puppetlabs.kitchensink.core :as ks]
-            [puppetlabs.services.jruby.jruby-puppet-schemas :as jruby-schemas]
-            [puppetlabs.services.jruby.jruby-puppet-internal :as jruby-internal]
-            [puppetlabs.services.jruby.jruby-puppet-agents :as jruby-agents]
+            [puppetlabs.services.jruby.jruby-schemas :as jruby-schemas]
+            [puppetlabs.services.jruby.jruby-internal :as jruby-internal]
+            [puppetlabs.services.jruby.jruby-agents :as jruby-agents]
             [clojure.java.io :as io]
             [clojure.tools.logging :as log])
-  (:import (puppetlabs.services.jruby.jruby_puppet_schemas JRubyPuppetInstance)
+  (:import (puppetlabs.services.jruby.jruby_schemas JRubyInstance)
            (clojure.lang IFn)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
@@ -40,12 +40,12 @@
 
 (schema/defn ^:always-validate
   get-pool :- jruby-schemas/pool-queue-type
-  "Gets the JRubyPuppet pool object from the pool context."
+  "Gets the JRuby pool object from the pool context."
   [context :- jruby-schemas/PoolContext]
   (jruby-internal/get-pool context))
 
 (schema/defn ^:always-validate
-  registered-instances :- [JRubyPuppetInstance]
+  registered-instances :- [JRubyInstance]
   [context :- jruby-schemas/PoolContext]
   (-> (get-pool context)
       .getRegisteredElements
@@ -67,14 +67,14 @@
 
 (schema/defn create-borrowed-event :- jruby-schemas/JRubyBorrowedEvent
   [requested-event :- jruby-schemas/JRubyRequestedEvent
-   instance :- jruby-schemas/JRubyPuppetBorrowResult]
+   instance :- jruby-schemas/JRubyBorrowResult]
   {:type :instance-borrowed
    :reason (:reason requested-event)
    :requested-event requested-event
    :instance instance})
 
 (schema/defn create-returned-event :- jruby-schemas/JRubyReturnedEvent
-  [instance :- jruby-schemas/JRubyPuppetInstanceOrPill
+  [instance :- jruby-schemas/JRubyInstanceOrPill
    reason :- jruby-schemas/JRubyEventReason]
   {:type :instance-returned
    :reason reason
@@ -113,12 +113,12 @@
 (schema/defn instance-borrowed :- jruby-schemas/JRubyBorrowedEvent
   [event-callbacks :- [IFn]
    requested-event :- jruby-schemas/JRubyRequestedEvent
-   instance :- jruby-schemas/JRubyPuppetBorrowResult]
+   instance :- jruby-schemas/JRubyBorrowResult]
   (notify-event-listeners event-callbacks (create-borrowed-event requested-event instance)))
 
 (schema/defn instance-returned :- jruby-schemas/JRubyReturnedEvent
   [event-callbacks :- [IFn]
-   instance :- jruby-schemas/JRubyPuppetInstanceOrPill
+   instance :- jruby-schemas/JRubyInstanceOrPill
    reason :- jruby-schemas/JRubyEventReason]
   (notify-event-listeners event-callbacks (create-returned-event instance reason)))
 
@@ -142,9 +142,9 @@
 ;;; Public
 
 (schema/defn ^:always-validate
-  initialize-config :- jruby-schemas/JRubyPuppetConfig
+  initialize-config :- jruby-schemas/JRubyConfig
   [config :- {schema/Keyword schema/Any}]
-  (-> (get-in config [:jruby-puppet])
+  (-> (get-in config [:jruby])
       (update-in [:compile-mode] #(keyword (or % default-jruby-compile-mode)))
       (update-in [:borrow-timeout] #(or % default-borrow-timeout))
       (update-in [:max-active-instances] #(or % (default-pool-size (ks/num-cpus))))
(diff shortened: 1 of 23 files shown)


src/clj/puppetlabs/services/jruby/jruby_internal.clj
```diff
@@ -1,16 +1,16 @@
-(ns puppetlabs.services.jruby.jruby-puppet-internal
+(ns puppetlabs.services.jruby.jruby-internal
   (:require [schema.core :as schema]
-            [puppetlabs.services.jruby.jruby-puppet-schemas :as jruby-schemas]
+            [puppetlabs.services.jruby.jruby-schemas :as jruby-schemas]
             [clojure.tools.logging :as log]
             [puppetlabs.kitchensink.core :as ks])
-  (:import (com.puppetlabs.puppetserver.pool JRubyPool)
-           (puppetlabs.services.jruby.jruby_puppet_schemas JRubyPuppetInstance PoisonPill ShutdownPoisonPill)
+  (:import (com.puppetlabs.jrubyutils.pool JRubyPool)
```
cprice404: I think it might be slightly more consistent to name this `com.puppetlabs.jruby_utils`, but I don't care very much :)

src/clj/puppetlabs/services/jruby/jruby_schemas.clj
```diff
@@ -72,35 +72,34 @@
...
-  {:config                JRubyPuppetConfig
+  "The data structure that stores all JRuby pools and the original configuration."
+  {:config                JRubyConfig
    :pool-agent            JRubyPoolAgent
    :flush-instance-agent  JRubyPoolAgent
    :pool-state            PoolStateContainer})
 
 (def JRubyInstanceState
-  "State metadata for an individual JRubyPuppet instance"
+  "State metadata for an individual JRuby instance"
```
cprice404: We don't actually have an object called just `JRuby`, right?  If not, these docstrings would probably be clearer if they said `JRubyInstance`.  (I recognize that they were also wrong before your changes, but still probably worth fixing.)

src/clj/puppetlabs/services/jruby/jruby_schemas.clj
```diff
@@ -72,35 +72,34 @@
...
   {:borrow-count schema/Int})
 
 (def JRubyInstanceStateContainer
-  "An atom containing the current state of a given JRubyPuppet instance."
+  "An atom containing the current state of a given JRuby instance."
```
cprice404: same

src/clj/puppetlabs/services/jruby/jruby_schemas.clj
```diff
@@ -40,19 +40,19 @@
...
-    * :ruby-load-path - a vector of file paths, containing the locations of puppet source code.
+    * :ruby-load-path - a vector of file paths, containing the locations of ruby source code.
 
     * :gem-home - The location that JRuby gems are stored
 
     * :compile-mode - The value to use for JRuby's CompileMode setting.  Legal
         values are `:jit`, `:force`, and `:off`.  Defaults to `:off`.
 
-    * :max-active-instances - The maximum number of JRubyPuppet instances that
+    * :max-active-instances - The maximum number of JRuby instances that
```
cprice404: same

src/clj/puppetlabs/services/jruby/jruby_schemas.clj
```diff
@@ -1,15 +1,15 @@
...
            (org.jruby Main Main$Status RubyInstanceConfig)
-           (com.puppetlabs.puppetserver.jruby ScriptingContainer)))
+           (com.puppetlabs.jrubyutils.jruby ScriptingContainer)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Schemas
 
 (def pool-queue-type
-  "The Java datastructure type used to store JRubyPuppet instances which are
+  "The Java datastructure type used to store JRuby instances which are
```
cprice404: same

src/clj/puppetlabs/services/jruby/jruby_internal.clj
```diff
@@ -133,14 +132,14 @@
...
-  [{size :max-active-instances} :- jruby-schemas/JRubyPuppetConfig]
+  [{size :max-active-instances} :- jruby-schemas/JRubyConfig]
   {:pool (instantiate-free-pool size)
    :size size})
 
 (schema/defn ^:always-validate
   cleanup-pool-instance!
-  "Cleans up and cleanly terminates a JRubyPuppet instance and removes it from the pool."
-  [{:keys [scripting-container pool] :as instance} :- JRubyPuppetInstance]
+  "Cleans up and cleanly terminates a JRuby instance and removes it from the pool."
```
cprice404: same

src/clj/puppetlabs/services/jruby/jruby_service.clj
```diff
@@ -68,20 +68,18 @@
     (let [event-callbacks (:event-callbacks (tk-services/service-context this))]
       (swap! event-callbacks conj callback-fn))))
 
-;; TODO: rename, get rid of references to Puppet
-(defmacro with-jruby-puppet
-  "Encapsulates the behavior of borrowing and returning an instance of
-  JRubyPuppet.  Example usage:
+(defmacro with-jruby-instance
+  "Encapsulates the behavior of borrowing and returning a JRuby instance.
```
cprice404: same

src/clj/puppetlabs/services/jruby/jruby_service.clj
```diff
@@ -68,20 +68,18 @@
...
+    (with-jruby-instance
+      jruby-instance
       jruby-service
-      (do-something-with-a-jruby-puppet-instance jruby-puppet)))
+      (do-something-with-a-jruby-instance jruby-instance)))
 
-  Will throw an IllegalStateException if borrowing an instance of
-  JRubyPuppet times out."
-  [jruby-puppet jruby-service reason & body]
+  Will throw an IllegalStateException if borrowing a JRuby instance times out."
```
cprice404: same

src/clj/puppetlabs/services/protocols/jruby.clj
```diff
@@ -1,11 +1,11 @@
-(ns puppetlabs.services.protocols.jruby-puppet)
+(ns puppetlabs.services.protocols.jruby)
 
-(defprotocol JRubyPuppetService
-  "Describes the JRubyPuppet provider service which pools JRubyPuppet instances."
+(defprotocol JRubyService
+  "Describes the JRuby provider service which pools JRuby instances."
```
cprice404: same

src/clj/puppetlabs/services/protocols/jruby.clj
```diff
@@ -26,7 +26,7 @@
 
   (free-instance-count
     [this]
-    "The number of free JRubyPuppet instances left in the pool.")
+    "The number of free JRuby instances left in the pool.")
```
cprice404: same

src/java/com/puppetlabs/jrubyutils/pool/JRubyPool.java
```diff
@@ -9,7 +9,7 @@
 import java.util.concurrent.locks.ReentrantLock;
 
 /**
- * An implementation of LockablePool for managing a pool of JRubyPuppet
+ * An implementation of LockablePool for managing a pool of JRuby
```
cprice404: same

src/clj/puppetlabs/services/jruby/jruby_core.clj
```diff
@@ -142,43 +142,43 @@
 ;;; Public
 
 (schema/defn ^:always-validate
-  initialize-config :- jruby-schemas/JRubyPuppetConfig
+  initialize-config :- jruby-schemas/JRubyConfig
   [config :- {schema/Keyword schema/Any}]
-  (-> (get-in config [:jruby-puppet])
+  (-> (get-in config [:jruby])
```
cprice404: We're going to have to be careful with this one, right?  If we change this, then it will be a breaking change in Puppet Server's config files?

cprice404: It seems like maybe what we might want to do is check first for the existence of a `:jruby` key in the config, and if it doesn't exist, then fall back to `:jruby-puppet`?  That way we have a path forward without breaking backward compat?

test/unit/puppetlabs/services/jruby/jruby_core_test.clj
```diff
@@ -12,7 +12,7 @@
 (def min-config
   {:product
    {:name "puppet-server", :update-server-url "http://localhost:11111"},
-   :jruby-puppet
+   :jruby
```
cprice404: Assuming we move forward with my suggestion above, about preferring a `:jruby` section but falling back to `:jruby-puppet` if it doesn't exist, then I think you should leave all of your test changes as they are in this PR (i.e., test the new setting by default).  And then maybe we just add one or two unit-ish tests that validate that we will fall back to `:jruby-puppet` if `:jruby` doesn't exist?

test/unit/puppetlabs/services/jruby/jruby_service_test.clj
```diff
@@ -23,15 +23,15 @@
...
+   (jruby-testutils/jruby-config {:max-active-instances pool-size})))
 
 (def default-services
-  [jruby-puppet-pooled-service])
+  [jruby-pooled-service])
 
 (deftest test-error-during-init
   (testing
-   (str "If there is an exception while putting a JRubyPuppet instance in "
+   (str "If there is an exception while putting a JRuby instance in "
```
cprice404: `JRubyInstance`

src/clj/puppetlabs/services/jruby/jruby_schemas.clj
rlinehan: I meant to make a comment about this on the PR, but I completely forgot. There were a lot of references to `JRubyPuppet` in comments and docstrings, and I think I mostly just renamed them all to `JRuby`. But I can go back and change things that were `JRubyPuppet instance` to `JRubyInstance` instead?

cprice404:
+1 other than nits, plus figuring out what we want to do for backward compat with the settings.

src/clj/puppetlabs/services/jruby/jruby_schemas.clj
cprice404: yeah, that seems like it'd be an improvement IMO.

src/clj/puppetlabs/services/jruby/jruby_core.clj
cprice404: Hmmm... now that I think about it... if we end up getting rid of the TK service, then we won't be reading any config automagically, we'll just have functions where the caller needs to pass us the config map, and maybe it's not unreasonable to expect Puppet Server to transform that map before passing it to us.  So maybe the way you have this right now is the best thing to do for the moment, and we can revisit after we figure out the Puppet Server integration a little more?

src/clj/puppetlabs/services/jruby/jruby_internal.clj
rlinehan: I swear I tried that and IntelliJ complained at me, but it seems like it must have been complaining for a different reason, because that totally works. Agreed that that's more consistent and an easy change.

src/clj/puppetlabs/services/jruby/jruby_core.clj
rlinehan: We have a ticket for "Refactor[ing] jruby-puppet config..." (SERVER-1189), and so I had assumed that for that we would do some refactoring of how this config works and integrates with Puppet Server. Even if we don't end up doing that ticket, IMO it makes more sense to leave this as-is right now and then see what we have to do for Puppet Server integration.

rlinehan:
@cprice404 I believe I've addressed all your feedback other than the change of `:jruby-puppet` to `:jruby` for the settings.

camlow325:
We might want to make the error message thrown when a timeout occurs from `with-jruby-instance` less "Puppet Server" specific. Maybe just have it say "Attempt to borrow a JRubyInstance from the pool timed out".  From where we'll consume this in Puppet Server, the upstream exception could be caught / rethrown with the Puppet Server specific content in it.

src/clj/puppetlabs/services/jruby/jruby_core.clj
camlow325: Yeah, I'm good with leaving it how @rlinehan has it here.  I'm also assuming the API for this library, assuming we get rid of any TK tie-ins, would have "puppet" references removed where appropriate.

cprice404:
+1 when @camlow325 is

rlinehan:
@camlow325 updated

camlow325:
:+1:

PR #7, opened 2016-05-11
(TK-367) Add initialization and cleanup hooks

rlinehan:
Add a lifecycle map to the jruby config that includes hooks for initialization, initialization of environment variables, shutdown, and shutdown-on-error that get called in the appropriate places. Users of this
library will then be able to add their own custom initialization and cleanup logic via this map.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_agents.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_agents.clj
@@ -63,7 +63,7 @@
   "Flush a single JRubyInstance.  Create a new replacement instance
   and insert it into the specified pool."
   [pool-context :- jruby-schemas/PoolContext
-   instance :- JRubyInstance
+   instance :- jruby-schemas/JRubyInstanceSchema
    new-pool :- jruby-schemas/pool-queue-type
    new-id :- schema/Int
    config :- jruby-schemas/JRubyConfig]
@@ -196,7 +196,7 @@
   "Sends requests to the flush-instance agent to flush the instance and create a new one."
   [pool-context :- jruby-schemas/PoolContext
    pool :- jruby-schemas/pool-queue-type
-   instance :- JRubyInstance]
+   instance :- jruby-schemas/JRubyInstanceSchema]
   ;; We use a separate agent from the main `pool-agent` here, because there is a possibility for deadlock otherwise.
   ;; e.g.:
   ;; 1. A flush-pool request comes in, and we start using the main pool agent to flush the pool.  We do that by
--- a/src/clj/puppetlabs/services/jruby/jruby_agents.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_agents.clj
@@ -41,7 +41,8 @@
   [{:keys [pool-state lifecycle] :as pool-context} :- jruby-schemas/PoolContext
    config :- jruby-schemas/JRubyConfig]
   (let [pool (:pool @pool-state)
-        init-fn (:initialize lifecycle)]
+        init-fn (:initialize lifecycle)
+        initialize-env-variables-fn (:initialize-env-variables lifecycle)]
     (log/debug (str "Initializing JRubyInstances with the following settings:\n"
                     (ks/pprint-to-string config)))
     (try
@@ -50,7 +51,7 @@
           (let [id (inc i)]
             (log/debugf "Priming JRubyInstance %d of %d" id count)
             (jruby-internal/create-pool-instance!
-             pool id config (partial send-flush-instance! pool-context) init-fn)
+             pool id config (partial send-flush-instance! pool-context) init-fn initialize-env-variables-fn)
             (log/infof "Finished creating JRubyInstance %d of %d"
                        id count))))
       (catch Exception e
@@ -68,10 +69,11 @@
    new-id :- schema/Int
    config :- jruby-schemas/JRubyConfig]
   (let [init-fn (get-in pool-context [:lifecycle :initialize])
-        shutdown-fn (get-in pool-context [:lifecycle :shutdown])]
+        shutdown-fn (get-in pool-context [:lifecycle :shutdown])
+        initialize-env-variables-fn (get-in pool-context [:lifecycle :initialize-env-variables])]
     (jruby-internal/cleanup-pool-instance! instance shutdown-fn)
     (jruby-internal/create-pool-instance!
-     new-pool new-id config (partial send-flush-instance! pool-context) init-fn)))
+     new-pool new-id config (partial send-flush-instance! pool-context) init-fn initialize-env-variables-fn)))
 
 (schema/defn ^:always-validate
   pool-initialized? :- schema/Bool
@@ -93,7 +95,8 @@
         old-pool (:pool old-pool-state)
         old-pool-size (:size old-pool-state)
         init-fn (:initialize lifecycle)
-        shutdown-fn (:shutdown lifecycle)]
+        shutdown-fn (:shutdown lifecycle)
+        initialize-env-variables-fn (:initialize-env-variables lifecycle)]
     (log/info "Replacing old JRuby pool with new instance.")
     (reset! pool-state new-pool-state)
     (log/info "Swapped JRuby pools, beginning cleanup of old pool.")
@@ -106,7 +109,12 @@
           (try
             (jruby-internal/cleanup-pool-instance! instance shutdown-fn)
             (when refill?
-              (jruby-internal/create-pool-instance! new-pool id config (partial send-flush-instance! pool-context) init-fn)
+              (jruby-internal/create-pool-instance! new-pool
+                                                    id
+                                                    config
+                                                    (partial send-flush-instance! pool-context)
+                                                    init-fn
+                                                    initialize-env-variables-fn)
               (log/infof "Finished creating JRubyInstance %d of %d"
                          id old-pool-size))
             (finally
--- a/src/clj/puppetlabs/services/jruby/jruby_agents.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_agents.clj
@@ -38,11 +38,9 @@
   prime-pool!
   "Sequentially fill the pool with new JRubyInstances.  NOTE: this
   function should never be called except by the pool-agent."
-  [{:keys [pool-state lifecycle] :as pool-context} :- jruby-schemas/PoolContext
+  [{:keys [pool-state] :as pool-context} :- jruby-schemas/PoolContext
    config :- jruby-schemas/JRubyConfig]
-  (let [pool (:pool @pool-state)
-        init-fn (:initialize lifecycle)
-        initialize-env-variables-fn (:initialize-env-variables lifecycle)]
+  (let [pool (:pool @pool-state)]
     (log/debug (str "Initializing JRubyInstances with the following settings:\n"
                     (ks/pprint-to-string config)))
     (try
@@ -50,8 +48,8 @@
         (dotimes [i count]
           (let [id (inc i)]
             (log/debugf "Priming JRubyInstance %d of %d" id count)
-            (jruby-internal/create-pool-instance!
-             pool id config (partial send-flush-instance! pool-context) init-fn initialize-env-variables-fn)
+            (jruby-internal/create-pool-instance! pool id config
+                                                  (partial send-flush-instance! pool-context))
             (log/infof "Finished creating JRubyInstance %d of %d"
                        id count))))
       (catch Exception e
@@ -68,12 +66,10 @@
    new-pool :- jruby-schemas/pool-queue-type
    new-id :- schema/Int
    config :- jruby-schemas/JRubyConfig]
-  (let [init-fn (get-in pool-context [:lifecycle :initialize])
-        shutdown-fn (get-in pool-context [:lifecycle :shutdown])
-        initialize-env-variables-fn (get-in pool-context [:lifecycle :initialize-env-variables])]
+  (let [shutdown-fn (get-in pool-context [:config :lifecycle :shutdown])]
     (jruby-internal/cleanup-pool-instance! instance shutdown-fn)
-    (jruby-internal/create-pool-instance!
-     new-pool new-id config (partial send-flush-instance! pool-context) init-fn initialize-env-variables-fn)))
+    (jruby-internal/create-pool-instance! new-pool new-id config
+                                          (partial send-flush-instance! pool-context))))
 
 (schema/defn ^:always-validate
   pool-initialized? :- schema/Bool
@@ -90,13 +86,11 @@
    old-pool-state :- jruby-schemas/PoolState
    new-pool-state :- jruby-schemas/PoolState
    refill? :- schema/Bool]
-  (let [{:keys [config pool-state lifecycle]} pool-context
+  (let [{:keys [config pool-state]} pool-context
         new-pool (:pool new-pool-state)
         old-pool (:pool old-pool-state)
         old-pool-size (:size old-pool-state)
-        init-fn (:initialize lifecycle)
-        shutdown-fn (:shutdown lifecycle)
-        initialize-env-variables-fn (:initialize-env-variables lifecycle)]
+        shutdown-fn (get-in config [:lifecycle :shutdown])]
     (log/info "Replacing old JRuby pool with new instance.")
     (reset! pool-state new-pool-state)
     (log/info "Swapped JRuby pools, beginning cleanup of old pool.")
@@ -109,12 +103,8 @@
           (try
             (jruby-internal/cleanup-pool-instance! instance shutdown-fn)
             (when refill?
-              (jruby-internal/create-pool-instance! new-pool
-                                                    id
-                                                    config
-                                                    (partial send-flush-instance! pool-context)
-                                                    init-fn
-                                                    initialize-env-variables-fn)
+              (jruby-internal/create-pool-instance! new-pool id config
+                                                    (partial send-flush-instance! pool-context))
               (log/infof "Finished creating JRubyInstance %d of %d"
                          id old-pool-size))
             (finally
--- a/src/clj/puppetlabs/services/jruby/jruby_agents.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_agents.clj
@@ -38,8 +38,7 @@
   prime-pool!
   "Sequentially fill the pool with new JRubyInstances.  NOTE: this
   function should never be called except by the pool-agent."
-  [{:keys [pool-state] :as pool-context} :- jruby-schemas/PoolContext
-   config :- jruby-schemas/JRubyConfig]
+  [{:keys [pool-state config] :as pool-context} :- jruby-schemas/PoolContext]
   (let [pool (:pool @pool-state)]
     (log/debug (str "Initializing JRubyInstances with the following settings:\n"
                     (ks/pprint-to-string config)))
@@ -173,8 +172,8 @@
   send-prime-pool! :- jruby-schemas/JRubyPoolAgent
   "Sends a request to the agent to prime the pool using the given pool context."
   [pool-context :- jruby-schemas/PoolContext]
-  (let [{:keys [pool-agent config]} pool-context]
-    (send-agent pool-agent #(prime-pool! pool-context config))))
+  (let [{:keys [pool-agent]} pool-context]
+    (send-agent pool-agent #(prime-pool! pool-context))))
 
 (schema/defn ^:always-validate
   send-flush-and-repopulate-pool! :- jruby-schemas/JRubyPoolAgent
--- a/src/clj/puppetlabs/services/jruby/jruby_agents.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_agents.clj
@@ -65,8 +65,8 @@
    new-pool :- jruby-schemas/pool-queue-type
    new-id :- schema/Int
    config :- jruby-schemas/JRubyConfig]
-  (let [shutdown-fn (get-in pool-context [:config :lifecycle :shutdown])]
-    (jruby-internal/cleanup-pool-instance! instance shutdown-fn)
+  (let [cleanup-fn (get-in pool-context [:config :lifecycle :cleanup])]
+    (jruby-internal/cleanup-pool-instance! instance cleanup-fn)
     (jruby-internal/create-pool-instance! new-pool new-id config
                                           (partial send-flush-instance! pool-context))))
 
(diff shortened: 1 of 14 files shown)


src/clj/puppetlabs/services/jruby/jruby_core.clj
```diff
@@ -142,26 +142,37 @@
 ;;; Public
 
 (schema/defn ^:always-validate
+  initialize-lifecycle-fns :- jruby-schemas/LifecycleFns
+  [config :- (schema/maybe {schema/Keyword schema/Any})]
+  (-> config
+      (update-in [:initialize] #(or % identity))
+      (update-in [:shutdown] #(or % identity))
+      (update-in [:shutdown-on-error] #(or % (fn [f] (f))))
```
rlinehan: I wasn't entirely sure whether we actually want to have defaults for these three or whether they should always have to be provided, but I went with provide defaults for now, and curious whether other folks have opinions.

src/clj/puppetlabs/services/jruby/jruby_service.clj
```diff
@@ -20,12 +21,15 @@
...
     [this context]
-    (let [config (core/initialize-config (get-config))
+    (let [initial-config (get-config)
           service-id (tk-services/service-id this)
-          agent-shutdown-fn (partial shutdown-on-error service-id)]
+          agent-shutdown-fn (partial shutdown-on-error service-id)
+          config (core/initialize-config (assoc-in initial-config
+                                                   [:jruby :lifecycle :shutdown-on-error]
+                                                   agent-shutdown-fn))]
       (core/verify-config-found! config)
```
rlinehan: This is somewhat unrelated to this PR, but it seems like with the location of where this line of code is, it will never do anything? The initial commit comes from https://github.com/puppetlabs/puppetserver/pull/123/commits/0ef0afd9ecb00c8cc558a2efb1b894463e1a3ecc, but there's no test and I don't think there's any way that `initialize-config` can return nil or an empty map, which is what this function checks for).

jpinsonault: Yeah this seems like pointless validation, and if we wanted that kind of validation is should be in the return schema of `initialize-config` or somewhere

jpinsonault:
I this looks good to me (unsurprisingly)

src/clj/puppetlabs/services/jruby/jruby_agents.clj
```diff
@@ -62,13 +61,14 @@
...
   [pool-context :- jruby-schemas/PoolContext
-   instance :- JRubyInstance
+   instance :- jruby-schemas/JRubyInstanceSchema
    new-pool :- jruby-schemas/pool-queue-type
    new-id :- schema/Int
    config :- jruby-schemas/JRubyConfig]
-  (jruby-internal/cleanup-pool-instance! instance)
-  (jruby-internal/create-pool-instance! new-pool new-id config
-                                        (partial send-flush-instance! pool-context)))
+  (let [shutdown-fn (get-in pool-context [:config :lifecycle :shutdown])]
```
cprice404: Do you think we could rename this lifecycle function to `:cleanup`?  Or `:terminate`?  The term `shutdown` is kind of overloaded in TK so I was a little thrown off by these references to "shutdown" at first.

src/clj/puppetlabs/services/jruby/jruby_core.clj
```diff
@@ -142,26 +142,37 @@
 ;;; Public
 
 (schema/defn ^:always-validate
+  initialize-lifecycle-fns :- jruby-schemas/LifecycleFns
+  [config :- (schema/maybe {schema/Keyword schema/Any})]
```
cprice404: This can have a more strict schema than this, right?  We know what some of the required keys in there are?

src/clj/puppetlabs/services/jruby/jruby_core.clj
cprice404: This seems like a good way to do it IMO.

src/clj/puppetlabs/services/jruby/jruby_internal.clj
```diff
@@ -139,39 +146,42 @@
...
   cleanup-pool-instance!
   "Cleans up and cleanly terminates a JRubyInstance and removes it from the pool."
-  [{:keys [scripting-container pool] :as instance} :- JRubyInstance]
+  [{:keys [scripting-container pool] :as instance} :- jruby-schemas/JRubyInstanceSchema
+   shutdown-fn :- IFn]
   (.unregister pool instance)
-  ;; TODO: need to add support for a callback hook, so that consumers like
-  ;; puppet-server can do their own cleanup.
-  ;(.terminate jruby-puppet)
+  (shutdown-fn instance)
```
cprice404: This would end up being called `cleanup-fn` as well, I presume, if you agree with my suggestion above.

cprice404: Does this fn need to be public?

src/clj/puppetlabs/services/jruby/jruby_internal.clj
```diff
@@ -139,39 +146,42 @@
...
+  create-pool-instance! :- jruby-schemas/JRubyInstanceSchema
   "Creates a new JRubyInstance and adds it to the pool."
   [pool :- jruby-schemas/pool-queue-type
    id :- schema/Int
    config :- jruby-schemas/JRubyConfig
    flush-instance-fn :- IFn]
-  (let [{:keys [ruby-load-path gem-home compile-mode]} config]
+  (let [{:keys [ruby-load-path gem-home compile-mode lifecycle]} config
+        init-fn (:initialize lifecycle)
+        init-env-vars-fn (:initialize-env-variables lifecycle)]
```
cprice404: Didn't @camlow325 come up with some idea about how we could handle the environment vars that wouldn't require us to make it an official lifecycle fn?  Like, I think maybe he was suggesting that we just always do our default environment var initialization, and then document how a user can override it as part of their implementation of their `init` fn?  My memory is rusty but I feel like he'd come up with some fairly compelling argument for doing it that way.

rlinehan: Yep, he did, the problem is it doesn't work :(. It seems like as soon as you call `.runScriptlet` that it does something special to set the environment variables, and then while you can call `.setEnvironment` after that, it doesn't get plumbed all the way through. :/

src/clj/puppetlabs/services/jruby/jruby_schemas.clj
```diff
@@ -99,19 +106,22 @@
...
                'JRubyInstanceState))
 
+(def JRubyInstanceSchema
+  {:pool pool-queue-type
+   :id schema/Int
+   :max-requests schema/Int
+   :flush-instance-fn IFn
+   :state JRubyInstanceStateContainer
+   :scripting-container ScriptingContainer
+   schema/Keyword schema/Any})
```
cprice404: Did this end up being necessary because there's no way to use schema to check if something satisfies the `record`?  Or because the schema check for the record is too strict and doesn't tolerate the final `schema/Keyword schema/Any` thing?

As-is it seems like we'll need to be really diligent about making sure that the fields here match up with the fields in the Record if they change over time?  If so, and there's no way for us to build the schema off of the record definition itself, then maybe we could add a comment that mentions that you need to edit them simultaneously?  Dunno.

src/clj/puppetlabs/services/jruby/jruby_service.clj
cprice404: looks like y'all are correct to me.

test/integration/puppetlabs/services/jruby/jruby_internal_test.clj
```diff
@@ -34,7 +34,7 @@
           instance (jruby-internal/create-pool-instance! pool 0 config #())
           container (:scripting-container instance)]
       (try
-        (= RubyInstanceConfig$CompileMode/JIT
-           (.getCompileMode container))
+        (is (= RubyInstanceConfig$CompileMode/JIT
+            (.getCompileMode container)))
```
cprice404: ha, nice catch

test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
```diff
@@ -8,7 +8,6 @@
             [puppetlabs.trapperkeeper.testutils.bootstrap :as tk-bootstrap]
             [puppetlabs.trapperkeeper.core :as tk]))
 
-(use-fixtures :each jruby-testutils/mock-pool-instance-fixture)
```
cprice404: hoozah!

src/clj/puppetlabs/services/jruby/jruby_internal.clj
cprice404: Ah, gotcha.

camlow325: Yeah, I'm trying to remember what arguments I made for trying to avoid this call...

I think I was concerned about allowing the consumer to be able to have access to the full `ScriptingContainer` instance so that the consumer could coordinate whatever current values are available in it with what gets used for the environment variables.  The design on this PR appears to allow for this still since the full `ScriptingContainer` instance appears to be passed in on the callback.

The other thing I hadn't mentioned at the time - but seems important - is that the consumer might want to just have access to the full set of default environment variables as JRuby itself would first set them up on a container instance rather than only get a chance to reset them later after our default logic had filtered them out.  For this case, I also think that what is on this PR would be better than what I'd suggested previously.

Where the kinds of actions that the consumer can perform during this lifecycle function wouldn't technically be limited to just setting up environment variables, I wonder, though, if it might be better to name this lifecycle function differently.  Like maybe we could have:

-  an `initialize-scripting-container` lifecycle function which is for performing any changes to the `ScriptingContainer` before the jruby-utils code would do any setup that involves calling `.runScriptlet` (after which point any further changes may be ignored).

... and ...

- an `initialize-pool-instance` lifecycle function which is about producing the instance that goes into the pool (but could receive a `ScriptingContainer` instance that had already had some Ruby code executed within it).

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -236,90 +237,149 @@
...
+          (is (true? (set-constants-and-verify pool-context 1)))
+          (let [instance (jruby-protocol/borrow-instance jruby-service
+                                                         :initialization-and-cleanup-hooks-test)]
+            (is (= "FOO" (:foo instance)))
+            (jruby-protocol/return-instance jruby-service instance :initialization-and-cleanup-hooks-test))
+
+          (jruby-protocol/flush-jruby-pool! jruby-service)
+          ; wait until the flush is complete
+          (await (get-in context [:pool-context :pool-agent]))
+          (is (logged? #"Terminating FOO"))))))))
```
cprice404: Any particular reason that we're implementing this as a logging test?  They tend to be fragile; seems like in this case maybe we could create an `atom` and modify its value in the cleanup fn, and then make an assertion on that?

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -236,90 +237,149 @@
...
+      (logutils/with-test-logging
+       (tk-testutils/with-app-with-config
+        app
+        [jruby/jruby-pooled-service]
+        config
+        (let [jruby-service (tk-app/get-service app :JRubyService)
+              context (tk-services/service-context jruby-service)
+              pool-context (:pool-context context)]
+          ;; set a ruby constant in each instance so that we can recognize them
+          (is (true? (set-constants-and-verify pool-context 1)))
```
cprice404: We're not checking these constants anywhere later in this test, are we?  So do we need this line?

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -236,90 +237,149 @@
...
-         ;; The jruby return instance calls done within the previous
-         ;; check jrubies call may cause an instance to be in the process of
-         ;; being flushed when the server is shut down.  This ensures that
-         ;; the flushing is all done before the server is shut down - since
-         ;; that could otherwise cause an annoying error message about the
-         ;; pool not being full at shut down to be displayed.
-         (timed-await (:flush-instance-agent pool-context)))))))
+             pool-context (:pool-context context)]
+         ;; set a ruby constant in each instance so that we can recognize them
+         (is (true? (set-constants-and-verify pool-context 1)))
```
cprice404: I don't think this one is necessary either.

test/unit/puppetlabs/services/jruby/jruby_agents_test.clj
```diff
@@ -104,29 +105,28 @@
...
+        (tk-testutils/with-app-with-config
+         app
+         [jruby/jruby-pooled-service]
+         config
+         (let [jruby-service (tk-app/get-service app :JRubyService)
+               context (tk-services/service-context jruby-service)]
+           (jruby-protocol/flush-jruby-pool! jruby-service)
+           ; wait until the flush is complete
+           (await (get-in context [:pool-context :pool-agent]))
+           (is (logged? #"Hello from shutdown"))))))))
```
cprice404: Same question about using logging for this test.  Also, is this redundant with the ones in the `int` namespace?

test/unit/puppetlabs/services/jruby/jruby_service_test.clj
```diff
@@ -34,21 +30,20 @@
    (str "If there is an exception while putting a JRubyInstance in "
         "the pool the application should shut down.")
     (logging/with-test-logging
-     (with-redefs [jruby-internal/create-pool-instance!
```
cprice404: yay!

test/unit/puppetlabs/services/jruby/jruby_testutils.clj
```diff
@@ -58,36 +50,6 @@
...
-  [f]
-  (with-redefs
-    [jruby-internal/create-pool-instance! create-mock-pool-instance]
-    (f)))
-
-(defmacro with-mock-pool-instance-fixture
-  [& body]
-  `(let [f# (fn [] (do ~@body))]
-     (mock-pool-instance-fixture f#)))
-
```
cprice404: such hooray

cprice404:
Looking good, I'm mostly +1.  Main things are:

1. Can we use a term like `cleanup` to represent the cleanup function, so we don't get it confused with the TK shutdown stuff.
2. Hopefully fix those tests that rely on logging.
3. whatever resolution we come to about the environment var init, re: @camlow325 's comments.

The other stuff I mentioned seems more in the "nit" category, so, probably wouldn't block on those.

src/clj/puppetlabs/services/jruby/jruby_agents.clj
rlinehan: `:cleanup` sounds good to me

src/clj/puppetlabs/services/jruby/jruby_core.clj
rlinehan: yep, will change... although it does match with what `initialize-config`, below, does.

src/clj/puppetlabs/services/jruby/jruby_internal.clj
rlinehan: I'm assuming you mean `cleanup-pool-instance!`... yes, probably, same as `create-pool-instance!` I've been holding off on moving stuff around because I think there's some other stuff in the same vein we'll want to do for removing the TK service and adding a pool manage service, and since I feel like there's still a little uncertainty there as to where things will end up, I was thinking of holding off on any reorganization until we start work on that ticket.

cprice404: That's fine.

src/clj/puppetlabs/services/jruby/jruby_core.clj
```diff
@@ -142,26 +142,37 @@
...
+(schema/defn ^:always-validate
   initialize-config :- jruby-schemas/JRubyConfig
   [config :- {schema/Keyword schema/Any}]
   (-> (get-in config [:jruby])
       (update-in [:compile-mode] #(keyword (or % default-jruby-compile-mode)))
       (update-in [:borrow-timeout] #(or % default-borrow-timeout))
       (update-in [:max-active-instances] #(or % (default-pool-size (ks/num-cpus))))
-      (update-in [:max-requests-per-instance] #(or % 0))))
+      (update-in [:max-requests-per-instance] #(or % 0))
+      (update-in [:lifecycle] #(initialize-lifecycle-fns %))))
```
camlow325: nit - seems like could just do:

```clj
(update-in [:lifecycle] initialize-lifecycle-fns)))
```

src/clj/puppetlabs/services/jruby/jruby_schemas.clj
rlinehan: The schema check for the record is too strict and doesn't tolerate any additional keys associated onto the record. I tried doing something like `(assoc JRubyInstance schema/Keyword schema/Any)` and `(merge {schema/Keyword schema/Any})` but no incantation of anything l tried worked, unfortunately. Unless anyone has any other ideas, I'll add a comment.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
rlinehan: For some reason I though that this was necessary in order to verify that the instance(s) were up before trying to borrow, but I guess that's not actually true, and you're right that we're not calling the corresponding `verify-no-constants` so this shouldn't be necessary.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
rlinehan: I think we just basically copied this out of the master-termination-test https://github.com/puppetlabs/jruby-utils/pull/7/files#diff-5420d9f56ec9b26871e6cd8bdce73df9L132, but agreed that using an atom would be less fragile.

test/unit/puppetlabs/services/jruby/jruby_agents_test.clj
rlinehan: Hmmm... I guess this is redundant? I feel like in the future, assuming we really do remove the TK service, it would be good to change this to not rely on a tk service at all, just call the underlying functions, and then maybe the one in the `int` namespace would define its own service and use that? i.e. I'm thinking that ultimately this test would be a bit more unit-y and the one in the `int` namespace would be more integration-y. Don't know if that's actually worthwhile though.

src/clj/puppetlabs/services/jruby/jruby_internal.clj
rlinehan: I think `initialize-scripting-container` and `initialize-pool-instance` make sense as names to me... they're a bit long, but definitely more explanatory/make more sense.

src/clj/puppetlabs/services/jruby/jruby_schemas.clj
camlow325: When I run the following code in a REPL with the current definition of `JRubyInstance`....

```clj
(schema/check JRubyInstance
              (map->JRubyInstance
               {:pool (JRubyPool. 10)
                :id 1
                :max-requests 100
                :flush-instance-fn identity
                :state (atom {:borrow-count 1})
                :scripting-container (ScriptingContainer.
                                      LocalContextScope/SINGLETON)
                :foo "bar"}))
```
I get:

```
=> {:foo disallowed-key}
```
However, if I add a map allowing extra key/value pairs to the JRubyInstance definition, like this:

```
(schema/defrecord JRubyInstance
  [pool :- pool-queue-type
   id :- schema/Int
   max-requests :- schema/Int
   flush-instance-fn :- IFn
   state :- JRubyInstanceStateContainer
   scripting-container :- ScriptingContainer]
  {schema/Keyword schema/Any})
```
... and then run the same `schema/check` from above, I now get:

```
=> nil
```
With this, I also still get an error if a required key is missing.  For example:

```clj
(schema/check JRubyInstance
              (map->JRubyInstance
               {:pool (JRubyPool. 10)
                :id 1
                :flush-instance-fn identity
                :state (atom {:borrow-count 1})
                :scripting-container (ScriptingContainer.
                                      LocalContextScope/SINGLETON)
                :foo "bar"}))
```

... would produce ...

```
{:max-requests (not (integer? nil))}
```

This utilizes the `extra-key-schema?` arg to the `defrecord` macro - https://github.com/plumatic/schema/blob/schema-1.0.4/src/cljx/schema/core.cljx#L1112-L1114.

Maybe this would work as an alternative that would allow us to not have to have the separate schema?

test/unit/puppetlabs/services/jruby/jruby_agents_test.clj
camlow325: It does feel redundant with the `int` tests currently.  I'd be inclined to remove/consolidate these tests along with the ticket where we reorient how the current `JRubyService` functions are exposed - borrow, return, flush, etc.

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
```diff
@@ -103,10 +101,11 @@
 (deftest prime-pools-failure
   (let [pool-size 2
         config        (jruby-testutils/jruby-config {:max-active-instances pool-size})
-        pool-context  (jruby-core/create-pool-context config jruby-testutils/default-shutdown-fn)
+        pool-context (jruby-core/create-pool-context config)
         err-msg       (re-pattern "Unable to borrow JRubyInstance from pool")]
-    (with-redefs [jruby-internal/create-pool-instance! (fn [_] (throw (IllegalStateException. "BORK!")))]
-                 (is (thrown? IllegalStateException (jruby-agents/prime-pool! pool-context config))))
+   (is (thrown? IllegalStateException (jruby-agents/prime-pool!
```
camlow325: 👯 1 more `with-redefs` down...

camlow325:
Mostly 👍 from me as well.  Just a few cosmetic comments.  Sounds like we're agreed on my comment about how to name the initialization hooks for setting environment variables vs. creating a pool instance.  Curious what you think about my comment on a possible way to extend the `schema/defrecord` that might allow us to get rid of the extra `JRubyInstanceSchema`.

src/clj/puppetlabs/services/jruby/jruby_schemas.clj
rlinehan: Yes, that seems like exactly what we want! Tried it out and it works perfectly.

rlinehan:
@camlow325 @cprice404 I believe I've addressed all your feedback

cprice404:
+1

camlow325:
👍 :shipit:

PR #8, opened 2016-05-13
(maint) Update maintenance section of README

rlinehan:
Correct spelling of "maintenance". Change where maintenance section says to
file tickets to TK, which matches the "support" section. TK seems more
appropriate than SERVER since this is not a Puppet Server specific library.


justinstoller:
👍

PR #9, opened 2016-05-24
(TK-370) Add pool manager service

rlinehan:
Add a pool manager service that handles creating and flushing a pool. We have
some concerns about whether multiple threads creating JRubyInstances at once
may cause problems due to JRuby's global state. Adding this service provides
us the ability to coordinate the initialization and shutdown of JRubyInstances
should our concerns about thread safety prove founded.

Update the JRuby service to use this pool manager service. In the future, we
will remove the JRuby TK service and this library will only provide the pool
manager service.

Also remove some unused dependencies from various namespaces.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_pool_manager_service.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_pool_manager_service.clj
@@ -12,22 +12,5 @@
    [this config]
    (log/info "Initializing the JRuby service")
    (let [pool-context (core/create-pool-context config)]
-     (jruby-agents/send-prime-pool! pool-context
-                                    (partial pool-manager-protocol/flush-instance! this))
-     pool-context))
-
-  (flush-instance!
-   [this pool-context pool instance]
-   (jruby-agents/send-flush-instance! pool-context pool instance))
-
-  (flush-pool!
-   [this pool-context]
-   (jruby-agents/send-flush-and-repopulate-pool! pool-context))
-
-  (flush-pool-for-shutdown!
-   [this pool-context]
-   (let [on-complete (promise)]
-     (log/debug "Beginning flush of JRuby pools for shutdown")
-     (jruby-agents/send-flush-pool-for-shutdown! pool-context on-complete)
-     @on-complete
-     (log/debug "Finished flush of JRuby pools for shutdown"))))
+     (jruby-agents/send-prime-pool! pool-context)
+     pool-context)))
--- a/src/clj/puppetlabs/services/jruby/jruby_pool_manager_service.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_pool_manager_service.clj
@@ -1,0 +1,28 @@
+(ns puppetlabs.services.jruby.jruby-pool-manager-service
+  (:require [puppetlabs.trapperkeeper.core :as trapperkeeper]
+            [puppetlabs.services.protocols.pool-manager :as pool-manager-protocol]
+            [puppetlabs.services.jruby.jruby-core :as core]
+            [puppetlabs.services.jruby.jruby-agents :as jruby-agents]
+            [clojure.tools.logging :as log]))
+
+(trapperkeeper/defservice jruby-pool-manager-service
+                          pool-manager-protocol/PoolManagerService
+                         []
+  (create-pool
+   [this config]
+   (log/info "Initializing the JRuby service")
+   (let [pool-context (core/create-pool-context config)]
+     (jruby-agents/send-prime-pool! pool-context)
+     pool-context))
+
+  (flush-pool!
+   [this pool-context]
+   (jruby-agents/send-flush-and-repopulate-pool! pool-context))
+
+  (flush-pool-for-shutdown!
+   [this pool-context]
+   (let [on-complete (promise)]
+     (log/debug "Beginning flush of JRuby pools for shutdown")
+     (jruby-agents/send-flush-pool-for-shutdown! pool-context on-complete)
+     @on-complete
+     (log/debug "Finished flush of JRuby pools for shutdown"))))
--- a/src/clj/puppetlabs/services/jruby/jruby_pool_manager_service.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_pool_manager_service.clj
@@ -7,13 +7,18 @@
 
 (trapperkeeper/defservice jruby-pool-manager-service
                           pool-manager-protocol/PoolManagerService
-                         []
+                          []
   (create-pool
    [this config]
    (log/info "Initializing the JRuby service")
    (let [pool-context (core/create-pool-context config)]
-     (jruby-agents/send-prime-pool! pool-context)
+     (jruby-agents/send-prime-pool! pool-context
+                                    (partial pool-manager-protocol/flush-instance! this))
      pool-context))
+
+  (flush-instance!
+   [this pool-context pool instance]
+   (jruby-agents/send-flush-instance! pool-context pool instance))
 
   (flush-pool!
    [this pool-context]
--- a/test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
+++ b/test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
@@ -146,7 +146,7 @@
 
 ;; TODO: this test seems redundant to the one in puppetlabs.services.jruby.jruby-agents-test
 ;; Reevaluate when we remove the jruby TK service
-(deftest ^:integration flush-jruby-pool-with-pool-manager-test
+(deftest ^:integration flush-pool-test
   (testing "Flushing the pool results in all new JRubyInstances"
     (let [config (jruby-testutils/jruby-config {:max-active-instances 4
                                                 :borrow-timeout default-borrow-timeout})]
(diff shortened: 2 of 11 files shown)


test/unit/puppetlabs/services/jruby/jruby_testutils.clj
```diff
@@ -96,3 +101,14 @@
...
+(defn wait-for-jrubies-from-pool-context
+  "Wait for all jrubies to land in the pool"
+  [pool-context]
+  (let [num-jrubies (-> pool-context
+                        :pool-state
+                        deref
+                        :size)]
+    (while (< (count (jruby-core/registered-instances pool-context))
+              num-jrubies)
+      (Thread/sleep 100))))
```
cprice404: Since the instantiation of the JRuby instances should be much faster in this repo, we could probably get away with just having a `Thread/yield` here, and maybe reduce test runtime by a smidge?

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
```diff
@@ -144,31 +145,51 @@
...
-        ;; now the pool is flushed, so the constants should be cleared
-        (is (true? (verify-no-constants pool-context 4)))))))
+      [pool-manager/jruby-pool-manager-service]
+      (jruby-testutils/jruby-tk-config config)
+      (let [pool-manager-service (tk-app/get-service app :PoolManagerService)
+            pool-context (pool-manager-protocol/create-pool pool-manager-service config)]
+        ;; wait for all jrubies to be added to the pool
+        (jruby-testutils/wait-for-jrubies-from-pool-context pool-context)
+        (is (= 4 (.size (jruby-core/get-pool pool-context))))
+        (pool-manager-protocol/flush-pool-for-shutdown! pool-manager-service pool-context)
```
cprice404: Is this synchronous?  It seems like there could be a race here.

cprice404:
LGTM other than the question about the shutdown flush test being possibly racey.

test/integration/puppetlabs/services/jruby/jruby_pool_int_test.clj
rlinehan: I moved the logic of creating a dereferencing the `on-complete` promise into the service function (https://github.com/puppetlabs/jruby-utils/pull/9/files#diff-033413699cb1199b181088cb75de4967R27), so it is now synchronous. However, not entirely sure that it should be?

cprice404: I'm not sure that I can immediately think of a reason why it couldn't be synchronous, other than the fact that that seems a little inconsistent with the other flush function, which is async.

Did you have a particular motivation (other than testing) for making it synchronous?

rlinehan: It reduces the logic that a user of the service would have to write - instead of creating the promise and dereferencing it, they can just call the service function... assuming they want it to be synchronous, as the jruby service has it. I couldn't think of any reason why you wouldn't want that call to be synchronous, but I also don't know the backstory for why the service stop was that way.

cprice404: OK, sorry, hadn't fully grokked the comparison between what the old one was doing and what this one is doing.  I get it now, and I can't think of a compelling reason why this isn't better, so I'm good with it as-is.

cprice404:
+1

rlinehan:
Added a commit changing from `(Thread/sleep)` to `(Thread/yield)` in the new test helper.

camlow325:
@rlinehan @cprice404 The basic API looks good to me - although I think we may also want to add a `flush-instance!` function to the new Trapperkeeper service.  The work that we do to flush a single instance - terminating an old one, creating a new one to replace it, adding that instance to the pool - is something I think we'll also want to coordinate on an agent at some point.  Layering it now via the TK service rather than leaving callers to invoke `send-flush-instance!` outside of the TK service seems like the right way to go.

@cprice404 - off topic a bit maybe - but were you thinking that all of the work that Puppet Server does today to decide to flush a pool instance when a max requests count has been reached is "client-specific" logic that we want to keep in Puppet Server rather than dragging into jruby-utils.  So clients would have the freedom to decide what criteria to use (possibly not at all) to flush individual instances and just use the TK service in jruby-utils at the point they've made the decision to flush?

src/clj/puppetlabs/services/jruby/jruby_pool_manager_service.clj
```diff
@@ -0,0 +1,28 @@
+(ns puppetlabs.services.jruby.jruby-pool-manager-service
+  (:require [puppetlabs.trapperkeeper.core :as trapperkeeper]
+            [puppetlabs.services.protocols.pool-manager :as pool-manager-protocol]
+            [puppetlabs.services.jruby.jruby-core :as core]
+            [puppetlabs.services.jruby.jruby-agents :as jruby-agents]
+            [clojure.tools.logging :as log]))
+
+(trapperkeeper/defservice jruby-pool-manager-service
+                          pool-manager-protocol/PoolManagerService
+                         []
```
jpinsonault: alignment is off on the brackets

jpinsonault:
Looks good to me once @camlow325's comment is addressed

cprice404:
> @rlinehan @cprice404 The basic API looks good to me - although I think we may also want to add a flush-instance! function to the new Trapperkeeper service. The work that we do to flush a single instance - terminating an old one, creating a new one to replace it, adding that instance to the pool - is something I think we'll also want to coordinate on an agent at some point. Layering it now via the TK service rather than leaving callers to invoke send-flush-instance! outside of the TK service seems like the right way to go.

Without looking at the code again, I agree with the principle expressed here; if a consumer ever needs to know that there is an agent in the pool context at all, we're not abstracting the API at the right level.  If that means that we need to add another function to prevent consumers from needing to use a `send` for anything, then I agree that we should do that.

@camlow325 yes I think `max-requests` is puppetserver-specific logic that shouldn't be included in this library unless we come across another use case for it from another project in the future.

rlinehan:
Right now the `max-requests` logic is used in `return-to-pool` - https://github.com/puppetlabs/jruby-utils/blob/master/src/clj/puppetlabs/services/jruby/jruby_internal.clj#L253-L263... so it sounds like instead of just calling `return-to-pool`, puppetserver would wrap that function with the `max-requests` logic somehow, or `return-to-pool` would take another argument allowing for some custom logic to determine when to flush the instance and when to just return the instance?

IMO it seems like there is a use-case for something like `max-requests` from other projects using this - from my recollection it was something requested because it was available prior to the existence of puppet server (there was an Apache setting - not specific to puppet - that allowed for something similar), and given our experience with leaky ruby code, it seems like not a bad thing to have available for users to use, rather than having to duplicate that logic themselves.

camlow325:
> IMO it seems like there is a use-case for something like max-requests from other projects using this...

@rlinehan That's a fair point.  The thing I'm not sure about yet is whether consumers interested in utilizing the flushing behavior would most commonly want to use the generic request count metric that we use in Puppet Server or might instead want to use something more sophisticated (and more domain-specific).  We know that for Puppet Server, for example, that the request counter metric has some significant limitations in that some types of requests - like catalog ones - should probably be weighted much higher than "simpler" ones with no extensibility (and, therefore, a much smaller likelihood for producing memory leaks in specific user environments) like file* ones.

I think I'd be inclined to let this play out upfront as Puppet Server having its own logic for request threshold flushing and see if other consumers that might want to do something around this as well would want to go with the simple request count based approach or something else.  If we end up in a situation where multiple consumers want exactly the same flushing behavior, I'd agree that pulling it into jruby-utils directly would make sense.  I'm just not sure we know enough there yet.

All I was thinking for now is that at the spot where Puppet Server calls the `flush-instance-fn` when checking if the count has been exceeded (https://github.com/puppetlabs/jruby-utils/blob/47e065a/src/clj/puppetlabs/services/jruby/jruby_internal.clj#L261) that Puppet Server would be calling through to the jruby-utils TK pool manager's `flush-instance!` function.  And, for now, that the implementation of the jruby-utils TK function could just call through to `send-flush-instance!` (https://github.com/puppetlabs/jruby-utils/blob/47e065a/src/clj/puppetlabs/services/jruby/jruby_agents.clj#L192) like you did for the `flush-pool!` function.

camlow325:
And if we go the way I was thinking for the `flush-instance!`, I think we'd be gutting the logic for checking the max request count from jruby-utils' version of `return-to-pool`, thinning it out to basically just be a pass through to a `.releaseItem` call on the underlying JRubyPool.  Might be better to do this thinning out work on a different ticket/PR if we decide to go this way?

cprice404:
Agree, let's tackle max-requests separately.  I'm waffling.

rlinehan:
@camlow325 I've added the `flush-instance!` function. For where Puppet Server calls `flush-instance-fn`, that's actually passed into the instance when it is created https://github.com/puppetlabs/jruby-utils/blob/47e065a0735ffedbcaeb42d39c6b52fc819bc2f0/src/clj/puppetlabs/services/jruby/jruby_agents.clj#L50-L51, and is then available on the instance. So I changed `prime-pool!` to take flush function and use that when creating the instances.

rlinehan:
@camlow325 @cprice404 added a new commit based on our discussion this morning. Also filed https://tickets.puppetlabs.com/browse/TK-376 for moving keys to `:internal` and https://tickets.puppetlabs.com/browse/TK-377 for determining what to do about `max-requests`.

cprice404:
+1 again :)  @camlow325 ?

camlow325:
LGTM. :shipit:

PR #10, opened 2016-06-02
(TK-366) Remove JRuby TK service

rlinehan:
Remove the jruby-pool TK service and protocol. Instead of this service, users
of this library will use the `pool-manager-service` to create a pool context,
and then pass that pool context to the functions in the `jruby-core`
namespace, to borrow, return, and flush instances.


test/integration/puppetlabs/services/jruby/jruby_locking_test.clj
```diff
@@ -2,93 +2,99 @@
...
-    (jruby-testutils/wait-for-jrubies app)
-    (let [jruby-service (tk-app/get-service app :JRubyService)]
-      (testing "initial state of write lock is unlocked"
-        (is (can-borrow-from-different-thread? jruby-service))
-      (testing "with-lock macro holds write lock while executing body"
-        (jruby-service/with-lock jruby-service :with-lock-holds-lock-test
-          (is (not (can-borrow-from-different-thread? jruby-service)))))
-      (testing "with-lock macro releases write lock after exectuing body"
-        (is (can-borrow-from-different-thread? jruby-service)))))))
+   app
```
rlinehan: There's a lot of whitespace changes in this PR that got mixed in with everything. I tried to minimize that, but it got really hard. Sorry. :( Maybe I have my intellij settings wrong, or our standards are different from when this code was first written. I realized after the fact that probably I should have tried to do the whitespace changes first, but by the time I realized that it was kind of too late.

src/clj/puppetlabs/services/jruby/jruby_core.clj
```diff
@@ -204,11 +204,10 @@
   timeout. If the timeout runs out then nil will be returned, indicating that
   there were no instances available."
   [pool-context :- jruby-schemas/PoolContext
-   timeout :- schema/Int
```
cprice404: So... I agree that it's weird that there was a borrow timeout in the context and it was also passed in here, but, this change seems like it forces consumers to only be able to have a single borrow timeout that can ever be used.

I can imagine a case where a consumer might have different types of uses for the JRuby instances, with different priorities / urgencies / failure conditions.  So they might want to have a really long borrow timeout for some of those cases and a really short one for others, and this change precludes that.

Is the copy of the borrow timeout that is in the context used elsewhere?  If so, where?  Maybe it would make sense to remove it from the config instead of removing it from this signature?

Was there anything else that motivated this change, or was it just an inconsistency that you noticed?

I don't want to drag this work out so I'm probably ok with this as-is; we could always add in another arity for this fn in the future if the use case that I described above were to come to pass.  But I am at least interested in knowing why the timeout is in the config.

cprice404: Also if we do keep it like this then I think the docstring needs to be updated.

src/clj/puppetlabs/services/jruby/jruby_core.clj
```diff
@@ -183,6 +184,12 @@
   [jruby-instance :- (schema/pred jruby-schemas/jruby-instance?)]
   @(:state jruby-instance))
 
+(schema/defn register-event-handler
+  "Register the callback function by adding it to the event callbacks atom."
+  [event-callbacks :- Atom
+   callback-fn :- IFn]
+  (swap! event-callbacks conj callback-fn))
```
cprice404: This part of this namespace is our "public" API, right?  If so, do we expect users to be passing around the `event-callbacks` Atom?

src/clj/puppetlabs/services/jruby/jruby_core.clj
```diff
@@ -282,3 +289,50 @@
...
+    (with-jruby-instance
+      jruby-instance
+      pool-context
+      event-callbacks
+      reason
+
+      (do-something-with-a-jruby-instance jruby-instance)))
+
+  Will throw an IllegalStateException if borrowing a JRubyInstance times out."
+  [jruby-instance pool-context reason event-callbacks & body]
```
cprice404: Same comment here, `event-callbacks` seems like a weird thing to expect the user to pass in.

src/clj/puppetlabs/services/jruby/jruby_core.clj
```diff
@@ -282,3 +289,50 @@
...
+       (do
+         (return-to-pool pool-instance# ~reason (deref ~event-callbacks))
+         (recur (borrow-from-pool-with-timeout ~pool-context ~reason (deref ~event-callbacks))))
+       (let [~jruby-instance pool-instance#]
+         (try
+           ~@body
+           (finally
+             (return-to-pool pool-instance# ~reason (deref ~event-callbacks))))))))
+
+(defmacro with-lock
```
cprice404: Same here with `event-callbacks`.

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
```diff
@@ -99,6 +99,62 @@
...
+          init-fn (fn [instance] @pool-initialized? instance)
+          pool-size 1
+          config (jruby-testutils/jruby-config
+                  {:max-active-instances pool-size
+                   :lifecycle {:initialize-pool-instance init-fn}})
+          pool-context (jruby-core/create-pool-context config)]
+
+      ;; start a pool initialization, which will block on the `initialize-pool-instance`
+      ;; function's deref of the promise
+      (jruby-agents/send-prime-pool! pool-context)
```
cprice404: curious why you aren't using a service here and calling `create-pool`?  As-is it seems like this test relies on knowledge of what happens inside that function?

Also this makes me wonder if `create-pool-context` should be in `core` or somewhere else?  Do we expect users to call it?

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
```diff
@@ -99,6 +99,62 @@
...
+                                     :borrow-during-pool-init-test
+                                     []))]
+
+        (is (not (realized? borrow-instance)))
+
+        ;; deliver the promise, allowing the pool initialization to complete
+        (deliver pool-initialized? true)
+
+        ;; now the borrow can complete
+        (is (some? @borrow-instance))))))
```
cprice404: are you just using `some?` as a shorthand for `not nil?` here?

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
```diff
@@ -99,6 +99,62 @@
...
+
+(deftest borrow-while-no-instances-available-test
+  (testing "when all instances are in use, borrow blocks until an instance becomes available"
+    (let [pool-size 2
+          config (jruby-testutils/jruby-config
+                  {:max-active-instances pool-size})
+          pool-context (jruby-core/create-pool-context config)]
+      (jruby-agents/prime-pool! pool-context)
```
cprice404: same question here about using a service

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
```diff
@@ -99,6 +99,62 @@
...
+          (is (not (realized? borrow-instance)))
+
+          ;; return an instance to the pool
+          (jruby-core/return-to-pool
+           (first drained-instances)
+           :borrow-with-no-free-instances-test
+           [])
+
+          ;; now the borrow can complete
+          (is (some? @borrow-instance)))))))
```
cprice404: When the pool falls out of scope without being shut down, and without having all of its instances returned to it, what happens?  I feel like we had some tests in the past where not cleaning up the pool properly before letting the test exit meant that there were still threads doing things in the background after the test ran, and that could cause problems with subsequent tests.  I don't *think* that's the case here, but thought it was worth mentioning.  I'd probably consider it 'best practice' to do a blocking shutdown of any pool that you create in a test before exiting from the test block, but if there are a ton of tests that would need to be updated for that, I don't know that I'd block the PR on it.

I think we got away with this a little more easily with the old service because it had a `stop` lifecycle function that would automatically get called by TK when we exited the `with-app-with-config` block of the test, and that would shut the pool down.

Raises the question of whether or not the new pool manager service should have a `stop` lifecycle that tries to clean up pools that haven't been cleaned up by the time it runs... but that definitely seems like a separate chunk of work.

test/unit/puppetlabs/services/jruby/jruby_service_test.clj
```diff
@@ -31,7 +62,7 @@
        (try
          (bootstrap/with-app-with-config
           app
-          jruby-testutils/default-services
+          (conj jruby-testutils/default-services jruby-pooled-test-service)
```
cprice404: not sure I understand why adding the new service to this particular test is necessary?

cprice404: Also that new service doesn't seem to be used anywhere besides this?  I'm sure I'm just missing something.

rlinehan: This test is testing that if an error happens during init, that the tk app will shut down. Since the `pool-manager-service` doesn't have anything happen during the init, it seemed like the best way to have this test was to include a service that has an `init` that is constructed in the way we expect users of this library to do. But maybe this test shouldn't exist at all anymore?

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
rlinehan: yep

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
rlinehan: All the other tests in this namespace use `create-pool-context`, and so I was mostly just being consistent with that. 

I don't think `create-pool-context` is something we expect users to call. Right now `core` has a combination of public and private functions, and this is in the "public" section right now, but I think it might make sense to make `core` solely public functions (and move this somewhere else).

cprice404:
Finished first pass review of all the remaining commits.  Everything else looks good to me!

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
cprice404: It's probably worth a quick spike to see how hard it would be to port them to using `create-pool`.  If it doesn't seem too difficult then that sounds like a good idea to me.

I am also in favor of moving non-public functions out of `core` but we could ticket that as a separate task and not block this PR on it.

cprice404: And... if porting the tests to use `create-pool` does end up seeming like a lot of work then I think I'm OK with leaving them as-is for now.

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
cprice404: ok.  haven't seen that before, and in this case it seems like the assertion could be something a little more specific, like verifying that the result is a actually an instance, but that's not a big deal so if you'd rather keep them like this that's fine.

test/unit/puppetlabs/services/jruby/jruby_service_test.clj
cprice404: Oh, I get it now.  You needed this because previously the default service list had a service like this in it, and it doesn't anymore.  This seems fine for now.

cprice404:
Most of my comments are about things that are somewhat cosmetic, so, @rlinehan , I'm happy to leave it to your judgment as to which of them you think are worth dealing with right now... the only real hangup I have is about the event-callbacks stuff.  It seems like we could probably figure out a way to stash those inside of the pool context so that they aren't part of the public API that users need to consume.  Happy to talk through that more if you like.  Other than that I think I'm pretty much +1 on this PR.

src/clj/puppetlabs/services/jruby/jruby_core.clj
rlinehan: I think it makes sense to stash `event-callbacks` on the `pool-context` rather than having users pass it in. `borrow-from-pool-with-timeout` and `return-to-pool` both already require having the deref'ed callback atom passed in; I think it makes sense for consistency to also remove it as an argument to them.

Currently the only argument passed to `create-pool` is the config. Do we think that `event-callbacks` should be part of the config, or that `create-pool` should take a second argument? Having it be part of the config feels a little weird, but maybe not that weird?

src/clj/puppetlabs/services/jruby/jruby_core.clj
rlinehan: >Was there anything else that motivated this change, or was it just an inconsistency that you noticed?

Yeah, I just took it out because it seemed like an inconsistency and like it would be easier for users not to have to pass it in.

It's not used elsewhere - in puppet server we take it out of the config, stuff it into the context, and then take it out of the context to pass to this function in the service's `borrow-instance`.

IMO it seems like part of the config - it gets read out of the config file or the default is inserted via `initialize-config`, so removing it from the config after the fact seems weird to me.

I definitely agree there's a use case for it, even if we don't have that use case yet in puppetserver. I'm fine with reverting this change, adding another arity, or adding a `borrow-from-pool-with-custom-timeout` function. I think I would prefer either of the last two.

src/clj/puppetlabs/services/jruby/jruby_core.clj
rlinehan: In this case I do think it makes sense to pass it in, since it's used for updating the atom.

However, I am actually not entirely sure of the use case for this, and whether it's still needed. I have only a vague idea of what the use case is (having multiple services be able to register event handlers), but probably a more concrete use case would help me understand.

For now, it seems like users would pass the atom into `create-pool` in the `init` of whatever service they are using the pool from, and then if the wanted to register event handlers from multiple other services they would create a `register-event-handler` service function in that service, and then have other services use that? Or would it make more sense for the `pool-manager-service` to still have a `register-event-handler` service function and handle creation/adding to the pool-context of the `event-callbacks` atom itself?

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
rlinehan: Created TK-379 for reorganizing the namespaces in this library and moving non-public functions out of `core`.

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
rlinehan: I feel like at one point (maybe on another team or something) I was told or heard the recommendation to use `some?` instead of `not nil?`. But agreed that asserting that it's actually an instance is a better test.

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
rlinehan: Unfortunately I think there are now a lot of tests that would have to be updated now :(, since I updated everything that used the old tk service to use the pool manager service, and as you said, it no longer has a stop. I haven't seen any problems yet in these tests, but I do think the idea of having a `stop` lifecycle function (generally, not just for these tests) is something worth exploring. I added TK-380 for this.

src/clj/puppetlabs/services/jruby/jruby_core.clj
cprice404: I guess I was thinking that we'd still have a `register-event-callback` fn in `core`, and people would call that, passing in the pool and the callback.  So, not on the config.  Would that work?

src/clj/puppetlabs/services/jruby/jruby_core.clj
cprice404: I'd say let's keep the change, update the docstring, and maybe add a TODO or comment saying that we should consider adding another arity in the future?

src/clj/puppetlabs/services/jruby/jruby_core.clj
cprice404: I think this is the same basic conversation as what I was commenting on above, but yeah I think we should still have a `register-event-handler` fn that hides the implementation details (atom), but that it could probably just be a regular function that takes the pool as an argument, rather than needing to be a service function.

As far as use cases go, there is a service in pe-puppet-server-extensions that registers a callback and uses it to compute all of the JRuby metrics in PE.  We might have to make the OSS puppet server jruby service have a 'register-event-handler' service function that closes over its pool object, so that the pe service can call that.

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
cprice404: it's a nitpick, so whatever you want to do (including nothing) is fine with me.

test/unit/puppetlabs/services/jruby/jruby_pool_test.clj
cprice404: OK, thanks, fine with punting on this for this PR.

src/clj/puppetlabs/services/jruby/jruby_core.clj
rlinehan: Yep, I think that makes sense to me.

rlinehan:
@cprice404 I think I've responded to all your feedback.

jpinsonault:
👍 looks good to me

PR #12, opened 2016-06-08
(TK-377) Rename 'max-requests' to 'max-borrows'

rlinehan:
Previously, the `max-requests` setting was used to determine if/when to flush
a JRuby instance after it had handled a certain number of requests in Puppet
Server. This logic might be useful to other users of this library (not just
Puppet Server), but such users might not actually have "requests" causing
JRubyInstances to be used. This commit renames `max-requests` (the setting and
all internal references) to `max-borrows`, which better reflects what it
actually does.


cprice404:
+1 pretty trivial changes.  Also pulled down the branch and grepped for the old name, and ran tests.  LGTM.

PR #13, opened 2016-06-09
(TK-376) Add :internal key to pool context and instance

rlinehan:
Both the pool context and the instance contain keys that we don't want users
to mess with - such as the pool and flush agents and the `flush-instance-fn`.
This commit moves all keys we want to keep "private" in both the pool context
and the instance underneath an `:internal` key.

For the the pool context, the `:config` key remains public; for the instance,
the `:id` and `:scripting-container` remain public.

### DIFF
--- a/src/clj/puppetlabs/services/jruby/jruby_agents.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_agents.clj
@@ -94,7 +94,7 @@
    new-pool-state :- jruby-schemas/PoolState
    refill? :- schema/Bool]
   (let [{:keys [config]} pool-context
-        pool-state-atom (get-in pool-context [:internal :pool-state])
+        pool-state-atom (jruby-internal/get-pool-state-container pool-context)
         new-pool (:pool new-pool-state)
         old-pool (:pool old-pool-state)
         old-pool-size (:size old-pool-state)
--- a/src/clj/puppetlabs/services/jruby/jruby_agents.clj
+++ b/src/clj/puppetlabs/services/jruby/jruby_agents.clj
@@ -20,6 +20,14 @@
       (mod next-id pool-size)
       next-id)))
 
+(schema/defn get-pool-agent :- jruby-schemas/JRubyPoolAgent
+  [pool-context :- jruby-schemas/PoolContext]
+  (get-in pool-context [:internal :pool-agent]))
+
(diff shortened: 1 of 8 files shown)


cprice404:
+1

src/clj/puppetlabs/services/jruby/jruby_agents.clj
```diff
@@ -85,13 +93,14 @@
    old-pool-state :- jruby-schemas/PoolState
    new-pool-state :- jruby-schemas/PoolState
    refill? :- schema/Bool]
-  (let [{:keys [config pool-state]} pool-context
+  (let [{:keys [config]} pool-context
+        pool-state-atom (get-in pool-context [:internal :pool-state])
```
camlow325: Would it make sense to call `(jruby-internal/get-pool-state pool-context)` here instead?

rlinehan: `get-pool-state` returns the dereferenced atom, but I can use `get-pool-state-container`, and yeah doing that sounds like a good idea.

camlow325: Cool - makes sense.

camlow325:
👍 and can merge after the one cosmetic change with the use of `get-pool-state-container` is merged in, unless we want to hold for @jpinsonault to review first.

rlinehan:
@camlow325 updated.

camlow325:
Looks like Travis failed on one cell, for oraclejdk7 only, for the latest run on this PR.  From https://travis-ci.org/puppetlabs/jruby-utils/jobs/137349890:

~~~
lein test puppetlabs.services.jruby.jruby-agents-test

No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
~~~~

Doesn't seem like it would be related to the changes in this PR.  @rlinehan, you agree?

rlinehan:
Hmmm, I agree, but also it's failed twice in a row now (I already reran it once).

jpinsonault:
Well I'm 👍. Are we good with merging (re: travis)?

rlinehan:
@jpinsonault I believe we are. We're going to change some of the `awaits` to `timed-awaits` in the namespace that was timing out, but we decided to do that in another PR and not block this one on it.

jpinsonault:
Cool

PR #17, opened 2016-08-01
(SERVER-584) improve handling of environment vars

janelu2:
This PR adds a whitelisting capability to the handling of environment variables. This also adds an `environment-vars` key to the list of accepted keys in JRubyConfig so that this key can be used in a `puppetserver.conf` file. Keys that are added to the new `environment-vars` map will be reachable and visible by any Ruby code. This PR also adds tests to make sure that the environment variables are correctly stored and visible.

### DIFF
--- a/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
@@ -100,12 +100,11 @@
   environment variables to be visible to the Ruby code. This map is by default
   set to {} if the user does not specify it in the configuration file."
   [env :- jruby-schemas/EnvMap
-   gem-home :- schema/Str
    config :- (schema/pred map?)]
   (let [whitelist ["HOME" "PATH"]
         clean-env (select-keys env whitelist)]
     (merge (assoc clean-env
-                  "GEM_HOME" gem-home
+                  "GEM_HOME" (:gem-home config)
                   "JARS_NO_REQUIRE" "true"
                   "JARS_REQUIRE" "false")
            (:environment-vars config))))
@@ -115,7 +114,7 @@
   container. Currently it just sets the environment variables."
   [scripting-container :- jruby-schemas/ConfigurableJRuby
    config :- jruby-schemas/JRubyConfig]
-  (.setEnvironment scripting-container (managed-environment (get-system-env) (:gem-home config) config))
+  (.setEnvironment scripting-container (managed-environment (get-system-env) config))
   scripting-container)
 
 (schema/defn ^:always-validate
--- a/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
@@ -107,7 +107,7 @@
                   "GEM_HOME" (:gem-home config)
                   "JARS_NO_REQUIRE" "true"
                   "JARS_REQUIRE" "false")
-           (:environment-vars config))))
+           (clojure.walk/stringify-keys (:environment-vars config)))))
 
 (schema/defn ^:always-validate default-initialize-scripting-container :- jruby-schemas/ConfigurableJRuby
   "Default lifecycle fn for initializing the settings on the scripting
--- a/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
@@ -94,7 +94,11 @@
   uses. JARS_NO_REQUIRE was the legacy way to turn off jar loading but is
   being phased out in favor of JARS_REQUIRE.  As of JRuby 1.7.20, only
   JARS_NO_REQUIRE is honored.  Setting both of those here for forward
-  compatibility."
+  compatibility.
+
+  We also merge an environment-vars map with the config to allow for whitelisted
+  environment variables to be visible to the Ruby code. This map is by default
+  set to {} if the user does not specify it in the configuration file."
   [env :- jruby-schemas/EnvMap
    gem-home :- schema/Str
    config :- (schema/pred map?)]
@@ -129,12 +133,15 @@
 
 (schema/defn ^:always-validate
   initialize-config :- jruby-schemas/JRubyConfig
+  "Initialize keys with default settings if they are not given a value.
+  The config is validated after these defaults are set."
   [config :- {schema/Keyword schema/Any}]
   (-> config
       (update-in [:compile-mode] #(keyword (or % default-jruby-compile-mode)))
       (update-in [:borrow-timeout] #(or % default-borrow-timeout))
       (update-in [:max-active-instances] #(or % (default-pool-size (ks/num-cpus))))
       (update-in [:max-borrows-per-instance] #(or % 0))
+      (update-in [:environment-vars] #(or % {}))
       (update-in [:lifecycle] initialize-lifecycle-fns)))
 
 (schema/defn register-event-handler
--- a/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
@@ -96,20 +96,22 @@
   JARS_NO_REQUIRE is honored.  Setting both of those here for forward
   compatibility."
   [env :- jruby-schemas/EnvMap
-   gem-home :- schema/Str]
+   gem-home :- schema/Str
+   config :- (schema/pred map?)]
   (let [whitelist ["HOME" "PATH"]
         clean-env (select-keys env whitelist)]
-    (assoc clean-env
-      "GEM_HOME" gem-home
-      "JARS_NO_REQUIRE" "true"
-      "JARS_REQUIRE" "false")))
+    (merge (assoc clean-env
+                  "GEM_HOME" gem-home
+                  "JARS_NO_REQUIRE" "true"
+                  "JARS_REQUIRE" "false")
+           (:environment-vars config))))
 
 (schema/defn ^:always-validate default-initialize-scripting-container :- jruby-schemas/ConfigurableJRuby
   "Default lifecycle fn for initializing the settings on the scripting
   container. Currently it just sets the environment variables."
   [scripting-container :- jruby-schemas/ConfigurableJRuby
    config :- jruby-schemas/JRubyConfig]
-  (.setEnvironment scripting-container (managed-environment (get-system-env) (:gem-home config)))
+  (.setEnvironment scripting-container (managed-environment (get-system-env) (:gem-home config) config))
   scripting-container)
 
 (schema/defn ^:always-validate
--- a/src/clj/puppetlabs/services/jruby_pool_manager/jruby_schemas.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/jruby_schemas.clj
@@ -61,7 +61,8 @@
     * :max-active-instances - The maximum number of JRubyInstances that
         will be pooled.
 
-    * :environment-vars - The whitelisted environment variables to be passed"
+    * :environment-vars - A map of whitelisted environment variables and their values to be
+        whitelisted and visible to any Ruby code."
   {:ruby-load-path [schema/Str]
    :gem-home schema/Str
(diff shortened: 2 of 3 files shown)


src/clj/puppetlabs/services/jruby_pool_manager/jruby_schemas.clj
```diff
@@ -59,14 +59,17 @@
         values are `:jit`, `:force`, and `:off`.  Defaults to `:off`.
 
     * :max-active-instances - The maximum number of JRubyInstances that
-        will be pooled."
+        will be pooled.
+
+    * :environment-vars - The whitelisted environment variables to be passed"
```
haus: This docstring is sort of confusing. When reading it, I think environment-vars will be a list of keys from the current environment to pass on to the jruby. However, the code below makes it look like it will be a set of predefined key/values.

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
```diff
@@ -96,20 +96,22 @@
   JARS_NO_REQUIRE is honored.  Setting both of those here for forward
   compatibility."
   [env :- jruby-schemas/EnvMap
```
haus: I think the docstring for this function needs to be updated to indicate arbitrary environment variables can be included when specified in the jruby config.

src/clj/puppetlabs/services/jruby_pool_manager/jruby_schemas.clj
janelu2: ok, would something like, "An optional map of whitelisted environment variables passed in from the puppetserver.conf file. These variables will be visible to Ruby code." be better?

haus: I'm interested in making it clear that the values will be passed in via the map and not from the outer environment.

janelu2: ohhhhh ok. that sounds good. i'll make that clear

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
janelu2: good point, i'll add something there.

src/clj/puppetlabs/services/jruby_pool_manager/jruby_schemas.clj
```diff
@@ -59,14 +59,17 @@
...
   {:ruby-load-path [schema/Str]
    :gem-home schema/Str
    :compile-mode SupportedJRubyCompileModes
    :borrow-timeout schema/Int
    :max-active-instances schema/Int
    :max-borrows-per-instance schema/Int
-   :lifecycle LifecycleFns})
+   :lifecycle LifecycleFns
+   (schema/optional-key :environment-vars) {schema/Str schema/Str}})
```
rlinehan: Right now everything in this config is required, and we don't validate this config until after we set a bunch of defaults - so things that are optional have a default in the `initialize-config` function (https://github.com/puppetlabs/jruby-utils/pull/17/files#diff-08581f068a6d2fb016f433b8ca45464eR131). This is kind of confusing and maybe not the best way of doing things, but I think for now it would be good for this new setting to be consistent - it should be required here and there should be a default in `initialize-config`.

janelu2:
@rlinehan fixed that stuff up and added some comments/better docstrings. Take a look sometime to make sure everything looks ok :)?

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
```diff
@@ -94,22 +94,28 @@
...
+                  "JARS_REQUIRE" "false")
+           (:environment-vars config))))
 
 (schema/defn ^:always-validate default-initialize-scripting-container :- jruby-schemas/ConfigurableJRuby
   "Default lifecycle fn for initializing the settings on the scripting
   container. Currently it just sets the environment variables."
   [scripting-container :- jruby-schemas/ConfigurableJRuby
    config :- jruby-schemas/JRubyConfig]
-  (.setEnvironment scripting-container (managed-environment (get-system-env) (:gem-home config)))
+  (.setEnvironment scripting-container (managed-environment (get-system-env) (:gem-home config) config))
```
camlow325: Just a style thing, but I might either:

- just pass in `config` and destructure the `:gem-home` and `:environment-vars` from within the `managed-environment` function 

or 

- destructure the config at this level and pass in the discrete values, like:

~~~clj
[{:keys [gem-home environment-vars]} :- jruby-schemas/JRubyConfig]
  (.setEnvironment scripting-container (managed-environment (get-system-env) gem-home environment-vars)
~~~

Pulling out one parameter but passing the whole config again as a second argument looks a little inconsistent to me.

Again, though, this is purely a cosmetic comment - no big deal.

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
```diff
@@ -94,22 +94,28 @@
...
-   gem-home :- schema/Str]
+   gem-home :- schema/Str
+   config :- (schema/pred map?)]
   (let [whitelist ["HOME" "PATH"]
         clean-env (select-keys env whitelist)]
-    (assoc clean-env
-      "GEM_HOME" gem-home
-      "JARS_NO_REQUIRE" "true"
-      "JARS_REQUIRE" "false")))
+    (merge (assoc clean-env
```
camlow325: @cprice404 I can't remember if we talked about our preferences for this before or not but do you think users should be able to use the custom environment variable configuration capability to override the special ones like `HOME`, `PATH`, and `GEM_HOME` that we define by default?  I think I'm okay with saying that they can do that, like @janelu2's implementation would do currently, so long as we document the ones that we have special values for so that users understand that they might cause problems if they choose to override our built-in defaults.  What do you think?

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
janelu2: Hahaha that is a good point. Do you feel strongly for one way or the other?

camlow325:
I'm basically 👍 on this.  I left the one question about override behavior in the case of conflicts that I'd be curious to get @cprice404's opinion about.  Otherwise all looks good to me.

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
camlow325: Whatever looks best to you would be fine with me.  Thanks!

janelu2:
Thanks @camlow325! I've made the change about having one `config` arg rather than that and an extracted key (lol). @cprice404 let me know your thoughts on jeremy's comment. Once this PR gets merged, the [puppetserver one](https://github.com/puppetlabs/puppetserver/pull/1161) will be able to start passing.

haus:
This will allow users to define their own env vars in configuration, but is there any plan to simply pass through a subset of env vars from the shell that launched the process?

rlinehan:
@haus if we made an environment-whitelist setting, that could add to the whitelist in `managed-environment`. Initially when making this library I had done something like that, but I believe we then decided that instead users could do that by providing their own `initialize-scripting-container-fn` which uses its own logic for determining what to do with environment variables, potentially using `managed-environment` since it's public (it's in the `jruby-core` namespace which is public API). It seems like it might be reasonable to have such a whitelist setting, but I'll admit I haven't read over the entirety of this ticket to know arguments for/against doing that. Just wanted to provide some background info.

src/clj/puppetlabs/services/jruby_pool_manager/jruby_schemas.clj
```diff
@@ -59,14 +59,18 @@
...
+        whitelisted and visible to any Ruby code."
   {:ruby-load-path [schema/Str]
    :gem-home schema/Str
    :compile-mode SupportedJRubyCompileModes
    :borrow-timeout schema/Int
    :max-active-instances schema/Int
    :max-borrows-per-instance schema/Int
-   :lifecycle LifecycleFns})
+   :lifecycle LifecycleFns
+   :environment-vars {schema/Str schema/Str}})
```
rlinehan: One thing to test out... I think that when the TK config service reads in HOCON files it usually makes keys in maps into clojure keywords. If that's the case, should this be keyword? Or would we require users to put things in the hocon file in double quotes (I think - although I'm not sure - that this would make them strings). It seems like it would be good to do a little manual testing of this.

rlinehan:
One comment about doing a little testing to make sure we know how people would have to specify these in the config. Otherwise 👍.

haus:
@rlinehan the main use cases i was thinking of are `.bashrc` files with complex logic for deciding what should be exported where that would be hard to do sanely in a config file. although, if hocon files can already use env vars, maybe they can just do something like:

`environment-vars: { "FOO": "${FOO}" }`

Where FOO is something defined in a `.bashrc`.

src/clj/puppetlabs/services/jruby_pool_manager/jruby_schemas.clj
janelu2: When I tried using keywords in the test, JRuby was giving me a lot of shit about it. I believe you have to have 2 strings. I might be mistaken.. @cprice404 ?

rlinehan: Sorry, I don't think I explained that clearly. I think that when passed into JRuby here they need to be strings. We should just double check that when we read off the config file that these can be strings, otherwise we'll need to convert what we get from the config file.

janelu2: oh! I gotcha. Should I do something like further up the page that calls a testing conf file?

rlinehan: I would start out by just testing it manually - maybe put something in the `environment-vars` setting in the config file in puppet server and running it out of the repl and seeing if that works.

janelu2: you were right, i hope this makes you happy

janelu2: :P @rlinehan I'm going to change them to keywords and see if everything breaks

rlinehan: If you put quotes around them in the config file does it work then?

camlow325:
@haus Yeah, that was exactly what @cprice404 was suggesting we would do for this feature.  Users would have to express any value which they want to derive from a shell environment variable in HOCON itself.  This is inspired by Nan Liu's suggestion from an old puppet-dev list thread [here](https://groups.google.com/d/msg/puppet-dev/Ac6gCTJXGnA/V25zx1fOtxMJ).  Given that, then, I think it's reasonable to have the work in the jruby-utils library operate on the literal string values for each variable and not have to have its own knowledge about the values of shell environment variables.

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
rlinehan: IMO the way this is currently seems reasonable/the simplest way of doing things, and trying to make it any more complicated would just be confusing. We already require `GEM_HOME` to be specified in the config, so sure, why not just let them override it if they really want to? We don't actually set anything specific for `HOME` and `PATH`, just let them be taken out of the environment variables. I do think it's worthwhile to document this though.

janelu2:
Just an update, I've changed the test maps and schema for JRubyConfig to take `{schema/keyword schema/str}` rather than the original `{schema/str schema/str}` since Hocon maps will be parsed to have this format. I've then added a `stringify-keys` call so that those keys change to strings when they are validated against EnvMap. 

Also I've verified that this works with an instance of puppetserver running in a repl.

janelu2:
@haus for what you said to work it should be defined 

`environment-vars: { "FOO": ${FOO} }`
not 
`environment-vars: { "FOO": "${FOO}" }`

but I just tested it and it's :+1:

haus:
@janelu2 cool. good to know it actually works. at some stage we should document the correct way to do that

camlow325:
LGTM.  :shipit:

PR #23, opened 2016-08-30
(TK-398) Fix with-jruby-retry-test race condition

camlow325:
This commit adds a borrow to the main service pool in the
`with-jruby-retry-test-via-mock-get-pool` function before any borrows
with intentional retries are attempted.  Previously, without the extra
borrow, the test's use of the `with-redefs` function to switch out the
`get-pool` function could cause pool instances to only be created in
the retry pool and not the main service pool, leading a borrow attempt
made against the main service pool to hang.

### DIFF
--- a/test/unit/puppetlabs/services/jruby_pool_manager/jruby_agents_test.clj
+++ b/test/unit/puppetlabs/services/jruby_pool_manager/jruby_agents_test.clj
@@ -53,6 +53,12 @@
      (let [config (jruby-testutils/jruby-config {:max-active-instances 1})
            pool-manager-service (tk-app/get-service app :PoolManagerService)
            pool-context (pool-manager-protocol/create-pool pool-manager-service config)]
+       (jruby-core/with-jruby-instance
+        jruby-instance
+        pool-context
+        :with-jruby-retry-test
+        (is (instance? JRubyInstance jruby-instance)
+            "Unable to borrow instance before using retry pool"))
        (let [real-pool (jruby-core/get-pool pool-context)
              retry-pool (JRubyPool. 1)
              _ (->> retry-pool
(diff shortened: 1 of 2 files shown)


test/unit/puppetlabs/services/jruby_pool_manager/jruby_agents_test.clj
```diff
@@ -68,7 +74,10 @@
             jruby-instance
             pool-context
             :with-jruby-retry-test
-            (is (instance? JRubyInstance jruby-instance))))
+            (is (instance? JRubyInstance jruby-instance)
+                (format
+                 "Unable to borrow instance after retries.  Num borrows: %d"
+                 @num-borrows))))
```
rlinehan: One thing that I noticed in the test that hung before is that the default borrow timeout is 20 minutes, but also the jenkins job will timeout if it doesn't receive output for 20 minutes. This is a bigger thing than just this test, but maybe it would be a good idea to lower the default borrow timeout in the test config?

rlinehan:
@camlow325 this looks good and like it should fix the issue... I left one question about maybe changing the default borrow time for tests, but that's not just a problem specific to this test/PR. 👍

test/unit/puppetlabs/services/jruby_pool_manager/jruby_agents_test.clj
camlow325: @rlinehan I think that's a great idea, thanks.  I pushed a separate commit, 03b0638, which sets the default borrow timeout in the `jruby-config` test helper to 5 minutes, hopefully shorter than a Jenkins or Travis timeout but plenty long enough for what the various tests in this suite need to do.

rlinehan:
💯

PR #24, opened 2016-09-15
(200) Add MAINTAINERS

cprice404:


cprice404:
@rlinehan I think we can go ahead and merge this one unless you think there are any other good candidates to put on the list.

rlinehan:
None that I can think of. lgtm.

PR #26, opened 2016-10-03
(SERVER-1412) Add support for `gem-path` config

cprice404:
This commit adds support for a `gem-path` configuration setting.
If specified it will be used as the value for the GEM_PATH
environment variable.  If not, GEM_PATH will be set to the same value
as GEM_HOME.

This is in support of SERVER-1412, where the server will pass in a value for this setting that includes a new "vendored" gem directory.  My intent is that Puppet Server will *always* pass in a value for this setting, and will have its own logic for defaulting the value that it uses for that setting.  But as far as the API of this library goes it seemed like this was the best way to build it.  Open to feedback - would be willing to make it a required argument if people thought that was better (that was my original instinct on how to implement this).

### DIFF
--- a/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
@@ -112,7 +112,6 @@
         clean-env (select-keys env whitelist)]
     (merge (-> (assoc clean-env
                  "GEM_HOME" (:gem-home config)
-
                  "JARS_NO_REQUIRE" "true"
                  "JARS_REQUIRE" "false")
                (add-gem-path config))
--- a/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
@@ -79,6 +79,13 @@
   []
   (into {} (System/getenv)))
 
+(schema/defn ^:always-validate add-gem-path
+  [env :- {schema/Str schema/Str}
+   config :- jruby-schemas/JRubyConfig]
+  (if-let [gem-path (:gem-path config)]
+    (assoc env "GEM_PATH" gem-path)
+    env))
+
 (schema/defn ^:always-validate managed-environment :- jruby-schemas/EnvMap
   "The environment variables that should be passed to the JRuby interpreters.
 
@@ -103,11 +110,12 @@
    config :- jruby-schemas/JRubyConfig]
   (let [whitelist ["HOME" "PATH"]
         clean-env (select-keys env whitelist)]
-    (merge (assoc clean-env
-                  "GEM_HOME" (:gem-home config)
-                  "GEM_PATH"  (:gem-path config)
-                  "JARS_NO_REQUIRE" "true"
-                  "JARS_REQUIRE" "false")
+    (merge (-> (assoc clean-env
+                 "GEM_HOME" (:gem-home config)
+
+                 "JARS_NO_REQUIRE" "true"
+                 "JARS_REQUIRE" "false")
+               (add-gem-path config))
            (clojure.walk/stringify-keys (:environment-vars config)))))
 
 (schema/defn ^:always-validate default-initialize-scripting-container :- jruby-schemas/ConfigurableJRuby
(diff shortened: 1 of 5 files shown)


src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
```diff
@@ -100,11 +100,12 @@
   environment variables to be visible to the Ruby code. This map is by default
   set to {} if the user does not specify it in the configuration file."
   [env :- jruby-schemas/EnvMap
-   config :- (schema/pred map?)]
+   config :- jruby-schemas/JRubyConfig]
```
cprice404: @rlinehan do you call whether there was a reason we were using the looser schema here, rather than an explicit `JRubyConfig`?  I'd be fine with switching it back but it seemed like we could actually get away with being more strict here, and the tests pass, so... :)

rlinehan:
Oh, didn't read all of the PR comment. I think having this be optional and default to `gem-home` make sense given how everything currently works with this library. Although I guess I don't fully understand the difference between `GEM_PATH` and `GEM_HOME` - is it common for them to be the same?

cprice404:
GEM_PATH should (AFAIK) always include GEM_HOME, but it may also include other paths.  GEM_PATH is where ruby will look when loading gems, GEM_HOME is where it will install new gems.

camlow325:
When I run `gem env` on a Puppet Server 2.6.0 install today, I see that JRuby has added in a couple of special GEM PATHS that it knows about:

```
  - GEM PATHS:
     - /opt/puppetlabs/server/data/puppetserver/jruby-gems
     - /root/.gem/jruby/1.9
     - file:/opt/puppetlabs/server/apps/puppetserver/puppet-server-release.jar!/META-INF/jruby.home/lib/ruby/gems/shared
```

In this case, the top one comes from `GEM_HOME`.  The `/root/.gem/jruby/1.9` seems ... undesirable?  The last one, though, is one I think we need to have in order for some of the built-in jruby gems to work properly.  With the changes on this PR, will that last one still be included even when we're explicitly setting the `GEM_PATH` variable?

cprice404:
@camlow325 I promise that I won't file a PR where jruby's stdlib isn't included in the path :)  Our tests  wouldn't pass anyway, right?

camlow325:
@cprice404, The thing I was wondering is whether if the GEM_PATH is explicitly set that the caller has to add on the stdlib directory in order for it to be there or if JRuby appends its stdlib path on internally.

If JRuby adds it automatically, seems like there's nothing we need to do.

If JRuby doesn't add it automatically, I'm wondering if we should have extra code in `jruby-utils` add the path in on behalf of the caller - so that we don't have to repeat that code in every other client.  Or, if that seems like a bad idea because we would want to give callers more control over whether or not the stdlib is added / or the specific order in which it is added, would it be better to have the "default" in jruby-utils for GEM_PATH to just not set GEM_PATH at all (rather than defaulting it to GEM_HOME)?  In that case, callers that don't need to customize the GEM_PATH would hopefully just get the default JRuby behaviors.

Does that make sense?

cprice404:
yeah, it does.

cprice404:
@camlow325 thinking about this a bit more - I think I like your suggestion of leaving GEM_PATH empty if it's not passed in here, regardless of the answer to the question about how `gem env` looks in my subsequent Puppet Server PR.  I think I'm going to go ahead and change that in this PR.

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
```diff
@@ -100,13 +107,15 @@
...
   (let [whitelist ["HOME" "PATH"]
         clean-env (select-keys env whitelist)]
-    (merge (assoc clean-env
-                  "GEM_HOME" (:gem-home config)
-                  "JARS_NO_REQUIRE" "true"
-                  "JARS_REQUIRE" "false")
+    (merge (-> (assoc clean-env
+                 "GEM_HOME" (:gem-home config)
+
```
camlow325: Cosmetic - extra line.

camlow325:
👍 With just the one dumb cosmetic extra line thingy.

camlow325:
👍

PR #27, opened 2016-10-05
(MAINT) prep for 0.3.0 release

cprice404:


cprice404:
This is blocked on puppetlabs/trapperkeeper#257

project.clj
```diff
@@ -34,16 +34,9 @@
                  ;; 'uberjar-exclusions' example toward the end of this file.
                  [org.jruby/jruby-stdlib "1.7.26"]
 
-
-                 [org.clojure/tools.logging "0.3.1"]
```
camlow325: Looks like we'd get version `0.2.6` from clj-parent -- which is a lower version than what we would have had previously.  Not sure if the newer version was intentional / required here?

cprice404: Probably just the latest version available when we created this repo.  Would be good to update it in the parent project but doesn't seem like a blocker.

haus:
sam-thumbs-up

PR #28, opened 2016-10-15
(SERVER-1585) Add `compat-version` setting

rlinehan:
Add a `compat-version` JRuby setting to allow configuration of the
compatibility version. The valid options are 1.9 or 2.0. This is tolerated as
a string or a double (so we don't need to worry about conversions between
Hiera/NC and HOCON parsing). The default is 1.9.

### DIFF
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
@@ -12,14 +12,6 @@
            (com.puppetlabs.jruby_utils.jruby ScriptingContainer)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-;;; Definitions
-
-(def compat-version
-  "The JRuby compatibility version to use for all ruby components, e.g. the
-  master service and CLI tools."
-  (CompatVersion/RUBY1_9))
-
-;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Private
 
 (schema/defn ^:always-validate initialize-gem-path :- {schema/Keyword schema/Any}
@@ -41,16 +33,24 @@
     :force RubyInstanceConfig$CompileMode/FORCE
     :off RubyInstanceConfig$CompileMode/OFF))
 
+(schema/defn ^:always-validate get-compat-version
+  "Get the JRuby compatibility version to use for all ruby components, e.g. the
+  master service and CLI tools."
+  [compat-version :- jruby-schemas/SupportedJRubyCompatVersions]
+  (case compat-version
+    "1.9" (CompatVersion/RUBY1_9)
+    "2.0" (CompatVersion/RUBY2_0)))
+
 (schema/defn ^:always-validate init-jruby :- jruby-schemas/ConfigurableJRuby
   "Applies configuration to a JRuby... thing.  See comments in `ConfigurableJRuby`
   schema for more details."
   [jruby :- jruby-schemas/ConfigurableJRuby
    config :- jruby-schemas/JRubyConfig]
-  (let [{:keys [ruby-load-path compile-mode lifecycle]} config
+  (let [{:keys [ruby-load-path compile-mode lifecycle compat-version]} config
         initialize-scripting-container-fn (:initialize-scripting-container lifecycle)]
     (doto jruby
       (.setLoadPaths ruby-load-path)
-      (.setCompatVersion compat-version)
+      (.setCompatVersion (get-compat-version compat-version))
       (.setCompileMode (get-compile-mode compile-mode)))
     (initialize-scripting-container-fn jruby config)))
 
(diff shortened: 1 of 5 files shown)


test/integration/puppetlabs/services/jruby_pool_manager/jruby_internal_test.clj
```diff
@@ -30,11 +30,28 @@
...
       (try
         (is (= RubyInstanceConfig$CompileMode/JIT
             (.getCompileMode container)))
+        (is (.is2_0 (.getCompatVersion container)))
         (finally
           (.terminate container))))))
+
+(deftest get-compat-version-test
+  (testing "returns correct compat version for SupportedJrubyCompatVersions enum"
+    (is (not (.is2_0 (jruby-internal/get-compat-version "1.9"))))
```
jpinsonault: Could this be `(is (.is1_9...`?

Huh, looks like it's a little funky, from the jruby code for `is1_9`
`return this == RUBY1_9 || this == RUBY2_0;`

Maybe check that `is1_9` is true and `is2_0` is false too?

jpinsonault:
Looks good! Just had 2 minor questions/comments

test/integration/puppetlabs/services/jruby_pool_manager/jruby_internal_test.clj
```diff
@@ -30,11 +30,28 @@
...
+    (is (.is2_0 (jruby-internal/get-compat-version "2.0"))))
+  (testing "returns a valid CompileMode for all values of enum"
```
cprice404: s/CompileMode/CompatVersion

cprice404:
LGTM other than minor comments above.

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
```diff
@@ -145,6 +149,7 @@
   [config :- {schema/Keyword schema/Any}]
   (-> config
       (update-in [:compile-mode] #(keyword (or % default-jruby-compile-mode)))
+      (update-in [:compat-version] #(str (or % default-jruby-compat-version)))
```
camlow325: I played around with defining `compat-version: 2.0` in a HOCON config.  I found that it converted the value to just an integer of '2' in memory.  If we're doing the `str` here just as a convenience to allow callers to pass the original value from the config on through to jruby-utils without touching it and we want to, for convenience, allow for users to be able to put the value in the config with or without quotes, we may want to put a little extra code in here to allow for '2' to be treated as '2.0'.  Here's a hacky change I tried that seemed to work for that here:

```clj
(update-in [:compat-version]
  #(if % (str (if (integer? %) (double %) %)) default-jruby-compat-version))
```

jpinsonault: Huh. I feel like hocon shouldn't convert floats to ints if they've got the decimal point. Maybe that's a bug?

I tried it out with the fancy new hocon cli too, it's the same there
```
echo "foo: 2.0" | ./bin/hocon get foo
2
```

rlinehan: Ugh, that's annoying. Yeah, I think that it should be possible to have the value in the config file be either `2.0` or `"2.0"` because I don't want to need to have to worry about whether the NC will convert the value to a string or not depending on how users enter it. @camlow325 I'll do something like your suggestion.

test/unit/puppetlabs/services/jruby_pool_manager/jruby_pool_test.clj
```diff
@@ -46,6 +46,29 @@
...
+      ;; depending on how the setting is laid down in a HOCON file, it seems feasible that it might
+      ;; be a string or a double. We should tolerate either.
+      (is (= "1.9" (-> minimal-config
+                       (assoc :compat-version 1.9)
+                       (jruby-core/initialize-config)
+                       :compat-version)))
+      (is (= "2.0" (-> minimal-config
+                       (assoc :compat-version 2.0)
+                       (jruby-core/initialize-config)
+                       :compat-version))))
```
rlinehan: The schema error right now is

```clojure
clojure.lang.ExceptionInfo: Output of initialize-config does not match schema: {:compat-version (not (#{"1.9" "2.0"} "2"))}
```
which seems like it gives enough info IMO?

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
```diff
@@ -41,16 +33,24 @@
     :force RubyInstanceConfig$CompileMode/FORCE
     :off RubyInstanceConfig$CompileMode/OFF))
 
+(schema/defn ^:always-validate get-compat-version
+  "Get the JRuby compatibility version to use for all ruby components, e.g. the
+  master service and CLI tools."
+  [compat-version :- jruby-schemas/SupportedJRubyCompatVersions]
+  (case compat-version
+    "1.9" (CompatVersion/RUBY1_9)
+    "2.0" (CompatVersion/RUBY2_0)))
```
rlinehan: In this case I don't think it will give a more user-friendly error message, since it should never get here - there will either be a schema error from the `initialize-config` schema, or from the schema error. However, I agree that having a default case makes sense, in case the schema changes or anything like that.

rlinehan:
Alright, I made some updates for y'all's comments @cprice404 @jpinsonault @camlow325. The result is that we parse an integer `2` into `2.0` and then later on we throw an error if it's `2` and not `2.0`... maybe a better way would be to just have `2` as part of the enum schema and in the case statement? Or maybe it's fine as-is. IDK.

test/unit/puppetlabs/services/jruby_pool_manager/jruby_pool_test.clj
jpinsonault: Yeah that seems good as far as schema errors go

jpinsonault:
Seems cleaner to keep it the way you have it, and not muddy up the `supported-jruby-compat-versions` schema with integers

jpinsonault:
👍

PR #30, opened 2016-10-28
(SERVER-797) Extract an interface for ScriptingContainer

camlow325:
This commit extracts an interface for use with schema/type validation
with ScriptingContainers.  This helps establish the baseline methods
that external consumers of the library can expect to be present for any
ScriptingContainers that the library creates for them.  The interface
is named ScriptingContainer.  The corresponding implementation class is
renamed to InternalScriptingContainer.

### DIFF
--- a/src/java/com/puppetlabs/jruby_utils/jruby/ScriptingContainer.java
+++ b/src/java/com/puppetlabs/jruby_utils/jruby/ScriptingContainer.java
@@ -7,22 +7,6 @@ import org.jruby.runtime.Block;
 /**
  */
 public interface ScriptingContainer extends EmbedRubyInstanceConfigAdapter {
-    Object callMethod(Object receiver, String methodName, Object... args);
-    Object callMethod(Object receiver, String methodName,
-                      Block block, Object... args);
-    <T> T callMethod(Object receiver, String methodName, Class<T> returnType);
-    <T> T callMethod(Object receiver, String methodName, Object singleArg,
-                     Class<T> returnType);
-    <T> T callMethod(Object receiver, String methodName, Object[] args,
-                     Class<T> returnType);
-    <T> T callMethod(Object receiver, String methodName, Object[] args,
-                     Block block, Class<T> returnType);
-    <T> T callMethod(Object receiver, String methodName, Class<T> returnType,
-                     EmbedEvalUnit unit);
-    <T> T callMethod(Object receiver, String methodName, Object[] args,
-                     Class<T> returnType, EmbedEvalUnit unit);
-    <T> T callMethod(Object receiver, String methodName, Object[] args,
-                     Block block, Class<T> returnType, EmbedEvalUnit unit);
     Object callMethodWithArgArray(Object receiver,
                               String methodName,
                               Object[] args,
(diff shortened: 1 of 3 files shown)


cprice404:
nit for future reference: doing the rename of the existing class as a separate commit would make a PR like this easier to review.  no biggie though.

src/java/com/puppetlabs/jruby_utils/jruby/ScriptingContainer.java
```diff
@@ -1,35 +1,32 @@
...
+    <T> T callMethod(Object receiver, String methodName, Object[] args,
+                     Class<T> returnType);
+    <T> T callMethod(Object receiver, String methodName, Object[] args,
+                     Block block, Class<T> returnType);
+    <T> T callMethod(Object receiver, String methodName, Class<T> returnType,
+                     EmbedEvalUnit unit);
+    <T> T callMethod(Object receiver, String methodName, Object[] args,
+                     Class<T> returnType, EmbedEvalUnit unit);
+    <T> T callMethod(Object receiver, String methodName, Object[] args,
+                     Block block, Class<T> returnType, EmbedEvalUnit unit);
```
cprice404: What would you think about removing some of these overloads from the interface for now?  I'm probably ok with merging this as-is if you think that this is the way to go, but TBH most of these signatures have generally served to just make it harder to interop from Clojure and I'm not really aware of any real use cases that we have for them.  I'd probably be OK with ditching *all* of them and waiting until we run into a case where `callMethodWithArgArray` isn't sufficient, or just keeping one or two of these that we feel more confident we'll end up actually using.

But, just throwing that out for discussion.  Not opposed to merging the PR with all of them if you prefer going that route.

camlow325:
I thought about doing the rename and the introduction of the interface as separate commits.  I didn't do that, though, because I generally like to have existing tests all pass on each commit and it would feel weird to rename the client references to the old class to `InternalScriptingContainer` only to have them be renamed back to `ScriptingContainer` (now an interface) on the next commit.  Thought it was useful to show that existing clients wouldn't have to change at all as part of the rename of the implementation class since an interface with the same name slides right in to replace it.  Anyway, will try to keep that in mind going forward.

cprice404:
> I thought about doing the rename and the introduction of the interface as separate commits. I didn't do that, though, because ...

cool, no biggie

src/java/com/puppetlabs/jruby_utils/jruby/ScriptingContainer.java
camlow325: Yeah, good suggestion.  I went ahead and did that in the related [puppetserver PR commit](https://github.com/puppetlabs/puppetserver/commit/38edc026160fb608708845260976756edaa1866f).  In 7468280, I went ahead and removed all of the `callMethod` methods from the interface.

jpinsonault:
👍  looks good to me

jpinsonault:
@camlow325 what's with the jenkins commits? Is that intentional?

camlow325:
@jpinsonault Ah, hm.  Rebase -i gone awry probably.  I'll fix that up.  Thanks for catching that!

camlow325:
@jpinsonault k, I think I got it fixed up with another rebase at feefac2.  Let me know if you see anything which still looks off about it.

jpinsonault:
👍

PR #32, opened 2016-11-17
(SERVER-377) Allow proxy environment variables to pass through

rlinehan:
jruby-utils limits the environment variables that it passes through to the
jruby interpreters. Previously, the list of environment variables allowed to
pass through did not include any proxy-related variables. This was a problem
because in some environments these variables are necessary for installing
gems, and rubygems respects these variables.

This commit adds the environment variables `HTTP_PROXY`, `http_proxy`,
`HTTPS_PROXY`, `https_proxy`, `NO_PROXY`, and `no_proxy` to be passed through
to the jruby scripting container.

In most cases developer workstations and CI systems won't have these variables
set, and there is no good way (without adding additional libraries and doing
some amount of hackery) to mock them out for tests. Thus, no tests have been
added or modified to test this change. However, there will be acceptance tests (using
beaker) added in puppetserver to test that these environment variables are now
being set.

### DIFF
--- a/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
@@ -90,6 +90,10 @@
     (assoc env "GEM_PATH" gem-path)
     env))
 
+(def env-allowed-list
+  "A list of environment variables that are allowed to be passed through from the environment."
+  ["HOME" "PATH" "HTTP_PROXY" "http_proxy" "HTTPS_PROXY" "https_proxy" "NO_PROXY" "no_proxy" "bar"])
+
 (schema/defn ^:always-validate managed-environment :- jruby-schemas/EnvMap
   "The environment variables that should be passed to the JRuby interpreters.
 
@@ -107,13 +111,12 @@
   JARS_NO_REQUIRE is honored.  Setting both of those here for forward
   compatibility.
 
-  We also merge an environment-vars map with the config to allow for whitelisted
+  We also merge an environment-vars map with the config to allow for configured
   environment variables to be visible to the Ruby code. This map is by default
   set to {} if the user does not specify it in the configuration file."
   [env :- jruby-schemas/EnvMap
    config :- jruby-schemas/JRubyConfig]
-  (let [whitelist ["HOME" "PATH"]
-        clean-env (select-keys env whitelist)]
+  (let [clean-env (select-keys env env-allowed-list)]
     (merge (-> (assoc clean-env
                  "GEM_HOME" (:gem-home config)
                  "JARS_NO_REQUIRE" "true"
--- a/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
@@ -90,9 +90,13 @@
     (assoc env "GEM_PATH" gem-path)
     env))
 
-(def env-allowed-list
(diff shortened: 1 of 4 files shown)


test/unit/puppetlabs/services/jruby_pool_manager/jruby_interpreter_test.clj
```diff
@@ -8,15 +8,26 @@
...
           jruby-env (.runScriptlet jruby-interpreter "ENV")]
-      ; $HOME and $PATH are left in by `jruby-env`
+      ;; $HOME and $PATH are left in by `jruby-env`
+      ;; Note that other environment variables are allowed through (e.g. `HTTP_PROXY` -
+      ;; see jruby-core/env-allowed-list for the full list), but are not
+      ;; expected to show up in most environments.
+      ;; If this test starts failing because CI systems or local development
+      ;; environments do have these variables set, we may need to revisit
+      ;; this test.
       (is (= #{"HOME" "PATH" "GEM_HOME" "JARS_NO_REQUIRE" "JARS_REQUIRE" "RUBY"}
```
camlow325: I made a similar comment in the corresponding [puppetserver PR](https://github.com/puppetlabs/puppetserver/pull/1289#discussion_r88682219).  I think it might be good here to filter out the http / no proxy variables that might end up being passed through to JRuby now (if a user happens to have had them set on their system when running the tests) before doing the equality comparison.  That way, this and the related `jruby-configured-env-vars` test could pass either way.

jpinsonault:
👍  I can't think of a reason why one set of env vars would be let through, but the proxy ones wouldn't, so it doesn't seem terrible to not have a test for them specifically

camlow325:
I also did some manual functional testing of this with the gem subcommand in Puppet Server, running from source.  I setup a SquidMan HTTP proxy locally and used the SquidMan access logs and Wireshark to inspect network requests.  I ran the following command at the command line:

```
HTTP_PROXY=localhost:8080 lein gem list nokogiri --remote
```

From the current HEAD of the master branch in puppetserver, the command succeeded (returning a list of remote gems available), but I could see from logs that no proxy was actually used.  The command also succeeded even if the proxy server had not been started.

Using a build of jruby-utils from this PR, the command failed (returning an empty list) if the proxy server had not been started.  After starting the proxy server and trying again, the command succeeded (returning a list of remote gems available).  In each case, I could see from the logs that the gem command was making requests through the proxy that I had configured.

This all looks good to me, then, at a functional level.

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
```diff
@@ -90,6 +90,10 @@
     (assoc env "GEM_PATH" gem-path)
     env))
 
+(def env-allowed-list
+  "A list of environment variables that are allowed to be passed through from the environment."
+  ["HOME" "PATH" "HTTP_PROXY" "http_proxy" "HTTPS_PROXY" "https_proxy" "NO_PROXY" "no_proxy" "bar"])
```
rlinehan: lol apparently I had a stray `"bar"` here. oops.

rlinehan:
@camlow325 @jpinsonault added a commit making the test more robust.

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
camlow325: hah, can't believe we missed that in the reviews as well.

camlow325:
Looks like @jpinsonault was good with this too, so merging...

PR #34, opened 2016-12-09
(PE-13552) Refactor jruby pool to use one pool

jpinsonault:
This PR modifies the flush logic to avoid the need for two pools,
which would get swapped out when a flush occurred.

It also removes the pool-agent and relies on the locking the pool to resolve multiple flush requests happening at once. Without pool swapping, the retry poison pill isn't needed anymore either.

The flush-instance-agent now takes the form of a modify-instance-agent,
so that any operations which create or destroy a jruby must go through
this agent

----

This should not be merged yet. It hasn't been integrated with puppetserver yet, and causes some test failures there. But it's in a good enough state to review


src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -86,47 +87,67 @@
...
+          (catch Exception e
+            (.clear pool)
+            (.insertPill pool (PoisonPill. e))
+            (throw (IllegalStateException.
+                    "There was a problem creating a JRubyInstance for the pool."
+                    e)))))))
+  (log/info "Finished draining pool."))
+
+(schema/defn ^:always-validate
+  drain-pool!
```
jpinsonault: I thought of that name too haha, try to think of something better

test/integration/puppetlabs/services/jruby_pool_manager/jruby_pool_int_test.clj
```diff
@@ -289,67 +192,41 @@
...
-                       pool-context
-                       10)))
+           (future (jruby-core/flush-pool! pool-context))
 
-           ;; at this point, we still have the main flush in progress, waiting for us
-           ;; to release the two instances from the old pool.  we should also have
-           ;; caused a flush of one of the two instances in the new pool, meaning that
-           ;; exactly one of the two in the new pool should have the ruby constant defined.
-           (is (true? (check-jrubies-for-constant-counts pool-context 1 1)))
+           ;; at this point, the flush is blocked, waiting for us
```
jpinsonault: Yeah good idea

test/integration/puppetlabs/services/jruby_pool_manager/jruby_pool_int_test.clj
```diff
@@ -289,67 +192,41 @@
...
          ;; and finally, we return the last instance from the old pool
          (jruby-core/return-to-pool instance1
                                     :max-borrows-flush-while-pool-flush-in-progress-test
                                     [])
 
          ;; wait until the flush is complete
-         (is (true? (jruby-testutils/timed-await (jruby-agents/get-pool-agent pool-context)))
-             (str "timed out waiting for the flush to complete, stack:\n"
-                  (get-all-stack-traces-as-str))))
+         (is (jruby-testutils/wait-for-instances pool 4) "Timed out waiting for the flush to finish"))
```
jpinsonault: Yeah you're right that's a race condition. I think you're right that if we wait for the future to finish, it will return after the jrubies have been borrowed, which will ensure that the `wait-for-instances` call won't accidentally return until the pool has actually been refilled

project.clj
```diff
@@ -53,5 +53,4 @@
                                    [puppetlabs/trapperkeeper :classifier "test" :scope "test"]]}
              :testutils {:source-paths ^:replace ["test/unit" "test/integration"]}}
 
-  :plugins [[lein-parent "0.3.1"]]
-  )
+  :plugins [[lein-parent "0.3.1" :exclusions [org.clojure/clojure]]])
```
jpinsonault: Ah, I forgot I had this in here. I pulled down the master branch and tried to run the tests, and got pedantic errors before I made any changes. I'm assuming it's something funky with my lein parent plugin or something, just hadn't figured the problem out yet

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -86,47 +87,67 @@
...
-                        old-pool)]
-          (try
-            (jruby-internal/cleanup-pool-instance! instance cleanup-fn)
-            (when refill?
-              (jruby-internal/create-pool-instance! new-pool id config
-                                                    (partial send-flush-instance! pool-context))
-              (log/infof "Finished creating JRubyInstance %d of %d"
-                         id old-pool-size))
-            (finally
-              (.releaseItem old-pool instance false))))
```
jpinsonault: looks like it's still being used in the `jruby-internal/return-to-pool` function so that an instance won't be returned to the pool if an exception occurs while flushing it

camlow325: Yeah, that was related to the case I mentioned where we needed to ensure that a lock counter was decremented when an instance was released - something that doesn't apply anymore with the current logic in the `JRubyPool`.  Effectively, that call ends up being a no-op.  The `queueLock` is taken and released with no actions being performed while the `queueLock` is held because `returnToPool` has a value of `false`.  If that's the only other case where you saw this being used, I'd still be in favor of eliminating the extra `.releaseItem` overload.

jpinsonault:
@jb pushed new commits that I think address all your feedback
@rlinehan or @KevinCorcoran this is in a good enough state now that you wouldn't be wasting your time if you wanted to take a look at it

test/integration/puppetlabs/services/jruby_pool_manager/jruby_pool_int_test.clj
```diff
@@ -184,80 +159,8 @@
          (is (= 1 (.size flushed-pool)))
          (is (jruby-schemas/shutdown-poison-pill? (.borrowItem flushed-pool))))))))
 
-(deftest ^:integration hold-instance-while-pool-flush-in-progress-test
```
jpinsonault: I think `max-borrows-flush-while-pool-flush-in-progress-test` handles the situation you described now

project.clj
camlow325: Did you still intend to remove this - or decide it actually is needed?

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -43,10 +38,27 @@
...
+
+  Blocks until the agent has completed filling the pool"
+  [pool-context :- jruby-schemas/PoolContext
+   insert-fn :- IFn]
+  (let [on-complete (promise)
+        pool-state (jruby-internal/get-pool-state pool-context)
+        fill-fn (fn []
+                  (jruby-internal/fill-pool-with-poison-pills! pool-state insert-fn)
+                  (deliver on-complete true))]
+    (send-agent (get-modify-instance-agent pool-context) fill-fn)
```
camlow325: Why was it that you were thinking we need to do the poison pill filling on the modify-instance agent?

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -86,87 +105,101 @@
...
+        pool (:pool pool-state)
+        borrow-fn (fn [_] (jruby-internal/borrow-from-pool!*
+                           jruby-internal/borrow-without-timeout-fn
+                           pool))]
+    (.lock pool)
+    (try
+      (mapv borrow-fn (range pool-size))
+      (catch Exception e
+        (.clear pool)
+        (.insertPill pool (PoisonPill. e))
```
camlow325: Should this be one of the cases that would now call `fill-pool-with-poison-pills!`?  If we end up going with the alternative approach I suggested around the `insertPill` implementation, though, I don't think this would need to change.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
```diff
@@ -86,6 +86,29 @@
...
+
+(schema/defn fill-pool-with-poison-pills!
+  "Fill an empty pool with poison pills. Should always be run on the
+  modify-instance-agent. insert-fn should be a fn that inserts a poison
+  pill of some kind into the pool"
+  [pool-state :- jruby-schemas/PoolState
+   insert-fn :- IFn]
+  (let [pool (:pool pool-state)
+        pool-size (:size pool-state)]
+    (when (not (= pool-size (.remainingCapacity pool)))
```
jpinsonault: Hmm, since creating pills will always happen on the modify-instance-agent I'm not too worried about overlapping pill insertions. This should only happen when you're shutting down, and 2 shutdown requests coming in at the same time won't step on each other's toes, since the second request will be waiting in the `.lock` call until the pills have filled the queue.

But regardless, I think it would make more sense conceptually to do what you're saying: have one slot for a pill in the pool, and use it to make decisions. Since the pills are only for shutting down/failures, that seems good to me. Much less ambiguity about how the pills work, and none of this filling the pool with pills nonsense

And yeah, I agree that if there is a pill in the pool no one should be able to get anything back from a borrow besides that pill

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -86,87 +105,101 @@
...
+                                                  (partial jruby-internal/insert-poison-pill e))
           (throw (IllegalStateException.
-                  "There was a problem adding a JRubyInstance to the pool."
-                  e)))))
-    ;; Add a "RetryPoisonPill" to the pool in case something else is in the
-    ;; process of borrowing from the old pool.
-    (.insertPill old-pool (RetryPoisonPill. old-pool))))
+                  "There was a problem creating a JRubyInstance for the pool."
+                  e))))))
+  (log/info "Finished draining pool."))
```
camlow325: Maybe we should make the draining/refilling message conditional here based on what `refill?` is set to?

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
camlow325: k, yeah, I think that makes sense - like for an insertPill that might happen from pool initialization vs. shutdown.  For dual shutdowns, the pool lock would probably now protect against the need to do this.  Might be a moot point if we go with the alternative `insertPill` implementation that I suggested on a separate comment.  Just curious.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -86,87 +105,101 @@
...
+
+(schema/defn ^:always-validate
+  drain-and-refill-pool!
+  "Borrow and destroy all the jruby instances, optionally refilling the
+  pool with fresh jrubies. Locks the pool in order to drain it, but releases
+  the lock before destroying the instances and refilling the pool"
+  [pool-context :- jruby-schemas/PoolContext
+   pool-state :- jruby-schemas/PoolState
+   refill? :- schema/Bool]
+  (log/info "Draining JRuby pool.")
```
camlow325: Same comment here on distinguishing drain vs. refill in the log statement.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
jpinsonault: Yeah definitely. Now that you've brought it up I hope we end up going with the way you talked about. Seems like fewer edge cases in that world

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
camlow325: > In the lock call where we await the allRegisteredInQueue Condition if the pool isn't full, short-circuit the await either if there the pill is non-null or if the pool has been refilled.

Assuming we go with something like the above... since the insertion of a pill should mean that the pool is effectively dead, I wonder if it might be good to have the `lock` call throw an `Exception` - internally releasing the pool lock in the process - if the `lock` is short-circuited by the insertion of a pill into the pool.  The `lock` method already can throw an `InterruptedException` if the thread on which the `lock` call is made is interrupted - which should only happen during a prolonged shutdown.

I think it might be bad if the `lock` call were to succeed, for example, if called by file sync after a pill has been inserted.  For example, file sync might otherwise be granted a lock on a pool that was only partially initialized and update code on disk while a "good" borrowed instance is in the middle of compiling a catalog.  If an exception is thrown from `lock` early (because the pool is borked and we need to shut down) file sync could then presumably bail out on applying a code update as well.

@KevinCorcoran - would definitely like your sanity check on this bit.  If this isn't making any sense, maybe we can get together for a few minutes after the holidays and go over this in person.

test/integration/puppetlabs/services/jruby_pool_manager/jruby_pool_int_test.clj
```diff
@@ -173,63 +148,37 @@
...
+           pool (:pool pool-state)]
        ;; wait for all jrubies to be added to the pool
        (jruby-testutils/wait-for-jrubies-from-pool-context pool-context)
        (is (= 4 (.size (jruby-core/get-pool pool-context))))
        (jruby-core/flush-pool-for-shutdown! pool-context)
-       (let [flushed-pool (jruby-core/get-pool pool-context)]
-         ;; flushing the pool removes all JRubyInstances but causes a ShutdownPoisonPill
-         ;; to be added
-         (is (= 1 (.size flushed-pool)))
-         (is (jruby-schemas/shutdown-poison-pill? (.borrowItem flushed-pool))))))))
```
camlow325: Shouldn't we still be able to get `ShutdownPoisonPill` at this point?

camlow325: ah, n/m.  Didn't look far enough down the diff on this.

test/integration/puppetlabs/services/jruby_pool_manager/jruby_pool_int_test.clj
```diff
@@ -289,67 +235,45 @@
...
+             ;; Wait until instance2 is returned
+             (is (jruby-testutils/wait-for-instances pool 3) "Timed out waiting for instance2 to return to pool")
+
+             ;; and finally, we return the last instance we borrowed to the pool
+             (jruby-core/return-to-pool instance1
+                                        :max-borrows-flush-while-pool-flush-in-progress-test
+                                        [])
+
+             ;; wait until the flush is complete
+             (deref flush-future 10000 false)
```
camlow325: Might be good to explicitly assert that the `deref` didn't time out?

test/integration/puppetlabs/services/jruby_pool_manager/jruby_pool_int_test.clj
```diff
@@ -289,67 +235,45 @@
...
+             (is (not (.isLocked pool)))
+
+             (is (jruby-testutils/wait-for-instances pool 4) "Timed out waiting for the flush to finish"))))
+
+       ;; we should have 4 fresh instances with the constant.
```
camlow325: s/with the constant/without the constant?

test/integration/puppetlabs/services/jruby_pool_manager/jruby_pool_int_test.clj
```diff
@@ -376,7 +300,7 @@
 
          (jruby-core/flush-pool! pool-context)
          ; wait until the flush is complete
-         (await (jruby-agents/get-pool-agent pool-context))
+         (is (jruby-testutils/wait-for-instances (jruby-internal/get-pool pool-context) 1))
```
camlow325: Did the `await` not work here anymore?  Seems like `wait-for-instances` could return true before the flush has started, depending upon timing.

test/unit/puppetlabs/services/jruby_pool_manager/jruby_agents_test.clj
```diff
@@ -104,5 +36,21 @@
              pool-context (pool-manager-protocol/create-pool pool-manager-service config)]
          (jruby-core/flush-pool! pool-context)
          ; wait until the flush is complete
-         (jruby-testutils/timed-await (jruby-agents/get-pool-agent pool-context))
+         (jruby-testutils/wait-for-instances (jruby-core/get-pool pool-context) 1)
```
camlow325: Same comment here as in one of the tests above about how it seems like this step could complete before the pool has actually been flushed, depending upon timing.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -86,87 +105,101 @@
...
+        borrow-fn (fn [_] (jruby-internal/borrow-from-pool!*
+                           jruby-internal/borrow-without-timeout-fn
+                           pool))]
+    (.lock pool)
+    (try
+      (mapv borrow-fn (range pool-size))
+      (catch Exception e
+        (.clear pool)
+        (.insertPill pool (PoisonPill. e))
+        (throw (IllegalStateException.
```
camlow325: In other cases where the `IllegalStateException` is thrown, it's typically from the `send-agent` function, which internally wraps the function in the `shutdown-on-error` macro.  I think the only place this gets called from is `drain-and-refill-pool!` but not inside of one the `shutdown-on-error` macro.  An exception thrown from here, then, may not trigger the server to shutdown, although I think the server should be shutdown in that case.  Could you see if this is the case and look into adding a `shutdown-on-error` wrapper, if necessary?

test/integration/puppetlabs/services/jruby_pool_manager/jruby_pool_int_test.clj
jpinsonault: derp, yes it would

test/integration/puppetlabs/services/jruby_pool_manager/jruby_pool_int_test.clj
jpinsonault: Yeah, I need to figure out a better way to wait for the flush to finish. The flush function takes the lock to get the jrubies, then does the actual refilling happens on the agent thread asynchronously. I think we were talking about how making the flush completely synchronous might be a good idea. That would remove the ambiguity in the tests and just make it more predictable in general for consumers

The only weird thing would be that jrubies will be in the pool available for borrowing before the flush completes since it destroys and remakes them one by one. The flush would always finish though, since it just creates a set number of jrubies and doesn't care if some have been borrowed before it finishes

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
jpinsonault: ahh, thanks for explaining that. Will do

test/integration/puppetlabs/services/jruby_pool_manager/jruby_pool_int_test.clj
camlow325: I think we had the `await` in there because it should block until the flush action has been completely processed through the agent.  We knew that on returning from the `flush-pool!` call that the flush action had at least been sent to the agent.  Under the hood, I think the `await` sends another action to the agent and waits for that to be processed.  By the time the second one has gone through, the first one should have finished.  Seems like this should still be the case now for determining when the flush has been completed on the pool agent?

> I think we were talking about how making the flush completely synchronous might be a good idea.

From the perspective of the HTTP flush jruby-pool endpoint, I think I'm okay with just the processing down through `collect-all-jrubies` being guaranteed to complete but not the actual refilling of the pool.  Waiting for the pool to completely fill back up would make the HTTP calls last longer.  Seems good enough to me if the caller can at least be guaranteed that any subsequent use of JRubies from the pool will be against new ones.

jpinsonault: Oh yeah, that's a good point. I was thinking that if you awaited the modify-instance-agent it would have a chance to sneak in before the flush's `send` call, but that can't happen since the flush's `send` would have to have happened before the flush call returns. Yeah, forget what I said about changing the flush call then. Agreed that it shouldn't wait for the whole pool to fill up before returning

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
jpinsonault: So since the `shutdown-on-error` fn is stored in the agent, and now we want to use it outside the agent. I'm thinking maybe the shutdown fn should be moved into the pool context instead of the agent context, or maybe both places, so that it's accessible outside of an agent. That'd be in `jruby-pool-manager-core/create-pool-context`

Thoughts?

camlow325: Yeah, I think it might make sense to store it in both places - maybe as a peer to `:modify-instance-agent` in the map associated with the `:internal` key in the pool context and as state in the agent itself.  Seems convenient to keep it in the agent state so that the receiving function in the agent can use it.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
KevinCorcoran: Consider this sequence:
* File sync requests the lock, blocks
* The pool is flushed - this puts the poison pill into the pool, right?
* As suggested, file sync gets an Exception

That makes plenty of sense, and I can see why you'd like to throw an Exception in that case, but from file sync's perspective it would be better if it could simply keep blocking until the new pool was available and acquire a lock on it, instead of getting an Exception.  That's not a hard requirement - I think it's probably acceptable to say that a flush can prevent code from being sync'ed, but it is a bit unfortunate.  Would that be a new limitation?  Do we know what happens today in this situation?

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -86,87 +105,101 @@
...
-          old-pool-state (jruby-internal/get-pool-state pool-context)
-          old-pool (:pool old-pool-state)
-          old-pool-size (:size old-pool-state)]
-      (when-not (pool-initialized? old-pool-size old-pool)
-        (throw (IllegalStateException. "Attempting to flush a pool that does not appear to have successfully initialized. Aborting.")))
-      (.insertPill new-pool (ShutdownPoisonPill. new-pool))
-      (swap-and-drain-pool! pool-context old-pool-state new-pool-state false))
-    (finally
-      (deliver on-complete true))))
+  ;; be queued up waiting for the lock, which can't be granted until all the instances
```
KevinCorcoran: Is this the desired behavior?  (queueing subsequent flush requests if one is already in progress)

It seems like it would be better to simply drop the subsequent flush requests in that case; or, maybe enqueue a single additional request (if instances can become in-use while the first flush is still in progress - not sure if that is true or not) but drop any more than one.

I realize that this comment may be outside the scope of these changes.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
jpinsonault: If I'm reading your comment right, there might be some confusion.

The poison pill will only be inserted if there's an error that caused the flush (or an error during the flush), not if it's just a normal flush from the http api.

If it's a normal flush, and filesync tries to take the lock, it will block until the pool is back up. Since there's now only one pool, and there's now swapping of pools happening, we don't have to insert the retry pill in the old pool just in case someone has a reference to it.

If there's been some error and a pill has been inserted, then the server will be shutting down, and I imagine filesync will want to know about that at some point, whether it's through an exception during the `.lock` call or when it finds a `ShutdownPoisonPill` after a borrow

jpinsonault:
Pushed a couple commits with the majority of what we discussed in the comments about changing the poison pill behavior. Still the open question about how `.lock` should be handled when there's a pill

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
KevinCorcoran: Oh, in that case, I think it's 💯  to throw an Exception since the lock will never be available anyway.

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
```diff
@@ -192,15 +204,11 @@ public E borrowItemWithTimeout(long timeout, TimeUnit unit) throws
...
-        releaseItem(e, true);
-    }
-
-    @Override
-    public void releaseItem(E e, boolean returnToPool) {
         final ReentrantLock lock = this.queueLock;
         lock.lock();
         try {
-            if (returnToPool) {
+            // Do nothing if the item being returned is the poisonPill
```
KevinCorcoran: What if it's a different poison pill?  Presumably that's impossible, I suppose?

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
```diff
@@ -389,7 +398,8 @@ private void signalIfAllRegisteredInQueue() {
         // is active at a time - a caller of lock() that has just acquired
         // the pool lock but is waiting for all registered elements to be
         // returned to the queue.
-        if (registeredElements.size() == liveQueue.size()) {
+        if (registeredElements.size() == liveQueue.size() ||
+                poisonPill != null) {
```
KevinCorcoran: Why check `poisonPill` here?

jpinsonault: I've gone back and looked at what uses `allRegisteredInQueue.await()`, and it looks like it's only used in `.lock`. So I think there's no reason =)

I think I misunderstood what it was used for when I made the change, so thanks for pointing this out

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
jpinsonault: It probably should be, but right now it isn't impossible. Because right now, it would be possible to return a different poison pill if:
1. You insert a pill
2. Borrow an instance (the pill)
3. Insert a different pill, overwriting the first
4. Return the first pill

We'll have to decide how should we handle two successive calls to `.insertPill`

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
```diff
@@ -134,6 +140,9 @@ public E borrowItem() throws InterruptedException {
             do {
                 if (isPoolLockHeldByAnotherThread(currentThread)) {
                     poolNotLocked.await();
+                } else if (this.poisonPill != null){
+                    // Return the pill immediately if there is one
+                    item = poisonPill;
```
camlow325: I think it might be okay to just return the poison pill even if the pool is already locked - in which case this could move to being the first `if` block.  @jpinsonault, @KevinCorcoran - anything you can think of where this might pose a problem?

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
```diff
@@ -173,6 +182,9 @@ public E borrowItemWithTimeout(long timeout, TimeUnit unit) throws
                     }
                     remainingMaxTimeToWait =
                             poolNotLocked.awaitNanos(remainingMaxTimeToWait);
+                } else if (this.poisonPill != null){
+                    // Return the pill immediately if there is one
+                    item = poisonPill;
```
camlow325: Same question here about whether this could be done before the pool lock await above.

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
jpinsonault: We could ignore any subsequent calls to `insertPill` or log a warning. If a pill has been inserted then the server should be shutting down. We could probably ignore it, since the error that caused a pill to be inserted should be thrown or logged where it was caught. It's not the pool's job to handle that.

I think the exception held in the pill can be returned to the user through an http request though? Not sure if that changes things.

@camlow325 thoughts?

camlow325: Another option here might be to see if the item being returned is something different than the poison pill or one of the previously registered pool instances and, if not, throw an `IllegalArgumentException` or something.  Seems like returning something that was never in the pool in the first place should be considered invalid.  That said, though, not sure if the extra validation is worth it.  I think I'm good with this as it is.  WDYT, @KevinCorcoran?

> We'll have to decide how should we handle two successive calls to .insertPill

For that, I think my vote would be to keep the first pill and not replace it with a subsequent one.  For purposes of returning the pill to a caller when translating to an error for logging, I think it would be good to hold onto information pertaining to why the original shutdown was triggered.  One could make an argument for having `.insertPill` throw an exception if a second or further pill is inserted if we're not going to retain it.  I'd probably avoid doing that, though.  In a more sophisticated solution later we might decide to hold onto an ordered list of all of the pills that were inserted for troubleshooting/error/logging purposes. I think having `.insertPill` calls beyond the first essentially be a no-op would be okay and maybe give us some flexibility for later.

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
camlow325: That work is probably still pending, but I had been thinking that in the `lock()` method we would be throwing an exception from the `lock()` method -- and, if necessary, freeing the pool lock -- if a poison pill exists.  Inside of the current `lock()` code, we can get stuck in an `await` if the pool is not full:

```clj
while (registeredElements.size() != liveQueue.size()) {
  allRegisteredInQueue.await();
}
```
With the new pill logic, I think we need to have that condition variable signaled when a pill is inserted into the pool so that the `lock()` caller doesn't get stuck indefinitely.  I had presumed that's what the extra check above was for.  Is that not right?

For clarity, I think it might be good to rename `allRegisteredInQueue` to something which indicates that a pill can disrupt a caller who is waiting for that to be signalled.  Maybe something like `pillInsertedOrAllRegisteredInQueue`.  Ugh, horrible naming... :(

test/integration/puppetlabs/services/jruby_pool_manager/jruby_pool_int_test.clj
```diff
@@ -156,9 +156,8 @@
        (is (= 4 (.size (jruby-core/get-pool pool-context))))
        (jruby-core/flush-pool-for-shutdown! pool-context)
 
-       ;; flushing the pool removes all JRubyInstances but causes ShutdownPoisonPills
-       ;; to be added, so it should still have 4 instances.
-       (is (= 4 (.size pool)))
+       ;; flushing the pool should remove all JRubyInstances
+       (is (= 0 (.size pool)))
        ;; But they should all be shutdown poison pills
        (is (every? jruby-schemas/shutdown-poison-pill? (.getRegisteredElements pool)))))))
```
camlow325: Should this still be true with the poison pill rework?  Wasn't clear to me that a pill would end up being a registered element in the pool.

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
camlow325: If we don't already have a unit test for that, I think having one where we have a `.lock()` block on waiting for the pool to fill back up and have the insertion of poison pill interrupt the wait (presumably throwing an exception) would be a good one to write.

test/unit/puppetlabs/jruby_utils/lockable_pool_test.clj
```diff
@@ -584,7 +584,24 @@
...
+  (testing "when borrow is blocked, inserting a pill unblocks it"
+    (let [pool (create-populated-pool 1)
+          pill (str "I'm just a pill, yes I'm only a pill")
+          instance (.borrowItem pool)
+          borrow-promise (promise)
+          blocked-borrow (future (let [instance (.borrowItem pool)]
+                                   (deliver borrow-promise instance)))]
+      (is (= 0 (.size pool)))
+      ; Pool is empty and the borrow future is blocked
+      (is (not (realized? borrow-promise)))
```
camlow325: Seems like you could just return the borrowed instance from the `future` and check here that the `blocked-borrow` future has not been realized rather than creating a separate promise?  Sometimes we will use a promise in a future, though, so that we can ensure that we at least got into the future before we make any assertions, though, like this:

```clj
about-to-borrow-promise (promise)
blocked-borrow (future (do (deliver about-to-borrow-promise true) (.borrowItem pool)))
...
(is @about-to-borrow-promise)
(is (not (realized? blocked-borrow)))
(.insertPill pool pill)
(let [borrow-result (timed-deref blocked-borrow)]
  (is (identical? borrow-result pill)))
```

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
jpinsonault: I was thinking that throwing the exception would take care of getting stuck waiting for the lock, but you're right, if the pill is inserted after the await, then it'll deadlock. 

Well at least the name's explicit

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
jpinsonault: I think that it would be ok to return it even if the pool is locked, since receiving the pill indicates the pool is borked and will never be unborked, and since the pill is not a limited resource like the queue that has to be carefully handed out, and since getting the pill doesn't allow the consumer to cause any side effects in the pool anyways. 

I'm not even sure we need to lock the queue before returning the pill, but it seems good to keep all the logic inside the `try/finally` block

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
camlow325: > Oh, in that case, I think it's 💯 to throw an Exception since the lock will never be available anyway.

Yeah, exactly.  I think this is ultimately going to be an improvement.  I think today that if we have a situation where a pill is inserted into the pool before it has been initially filled up or around shutdown time and a file sync lock is attempted, the `lock()` call will probably be stuck indefinitely - presumably until the process is SIGKILLed by the OS service manager.  I was mostly just trying to gauge whether going with an exception vs. just granting the lock was preferable.  I think an exception is better, though.  If we know we want to shut down as soon as possible after this happens, I wouldn't think we'd want to allow a potentially long-running background task like a file sync code update to be kicked off.

In terms of what kind of exception we'd throw here, I think it would be good to have an identifiable one that receivers could implement custom handling for.  For example, in the case of the exception being thrown because Trapperkeeper has called stop twice on the JRuby service, it might be good to just eat a 'poison pill' exception rather than just put extra noise into the logs - since, per the point @jpinsonault made elsewhere - we should log statements around where we do `insertPill` calls anyway about why the pill is being inserted.  It might make sense to codify a new `PillException` class and have the actual pill be accessible through a getter method on the exception in the event that the caller would like to / need to inspect it for some reason.

For convenience, it might make sense to just have a `PillException` be a class extended from `InterruptedException`.  That way, callers that don't care about the difference between the `InterruptedException` - which many of the methods in the `JRubyPool` class already throw - and a `PillException` could handle both types of exception as the base type.  Semantically, I think one could argue that `PillException` is a specialized form of interruption which can occur if a pill is inserted into the pool while the caller is waiting for a borrow, lock, or whatever to complete.

Definitely open to other names / suggestions around the naming, though.

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
camlow325: > I'm not even sure we need to lock the queue before returning the pill, but it seems good to keep all the logic inside the try/finally block.

I think it would be good to have anything which gets/sets the `poisonPill` member variable be protected by some sort of synchronization mechanism since those actions could happen from different threads around the same time.  Since we already use the `queueLock` for that, seems like it would be okay to continue doing that here.

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
```diff
@@ -86,6 +86,12 @@
     // to impose any noticeable performance degradation.
     private volatile Thread poolLockThread = null;
 
+    // Holds a poison pill object for errors and shutdowns
+    // If not null, takes priority over any pool instance when a call to
+    // borrowItem is made. Returns made using releaseItem are ignored if the
+    // released item is the poison pill already stored here
+    private E poisonPill;
```
camlow325: I think we would want to mark this variable as `volatile` like we do the `poolLockThread` above since the value can be gotten / set from different threads concurrently.  @KevinCorcoran - that sound right to you?

Cosmetic suggestion - maybe it would be good to just call this `pill` like we do with the existing `insertPill` method?  We use the term 'poison pill' informally but with the different kinds of pills that can actually be inserted into the pool, might be better to just go with the more generic name everywhere.

test/unit/puppetlabs/jruby_utils/lockable_pool_test.clj
jpinsonault: Ah yep it's completely redundant isn't it. I'm not sure we'd need to create a `about-to-borrow-promise`, since after dereffing that we still aren't totally sure the `borrowItem` call is blocking yet anyways when we check that it's not realized. Not sure if we can guarantee that it's blocking before we insert the pill unfortunately, without sleeping or something

camlow325: Yeah, the `about-to-borrow-promise` is just a tactic for trying to make the code a bit less racy but definitely not perfect either.  Maybe just a short sleep (like 100 ms) without the promise would help increase the likelihood that we've gotten into the borrow before the pill is inserted without making the tests onerously slower to run.  Whatever you want to do there is fine with me.

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
KevinCorcoran: > Another option here might be to see if the item being returned is something different than the poison pill or one of the previously registered pool instances and, if not, throw an IllegalArgumentException or something. Seems like returning something that was never in the pool in the first place should be considered invalid. That said, though, not sure if the extra validation is worth it. I think I'm good with this as it is. WDYT, @KevinCorcoran?

This seems like a decent approach.  The extra validation is probably unnecessary from a functionality perspective, but it seems like it would make the code a little easier to understand - i.e. Future Us won't have to look at this and wonder "what happens if it's a different pill?"

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
KevinCorcoran: Yes, I was wondering if it needed to be volatile as well.  I think it does.

KevinCorcoran:
Moving this thread out of an outdated diff, because they are impossible to find:

> For convenience, it might make sense to just have a PillException be a class extended from InterruptedException. That way, callers that don't care about the difference between the InterruptedException - which many of the methods in the JRubyPool class already throw - and a PillException could handle both types of exception as the base type. Semantically, I think one could argue that PillException is a specialized form of interruption which can occur if a pill is inserted into the pool while the caller is waiting for a borrow, lock, or whatever to complete.

If we end up creating a new Exception class for this, I think it would be best to name it something which does not include the word "pill". As file sync, I don't care at all that the pool uses poison pills - that's an implementation detail, so leaking the word "pill" into the API feels wrong.  The alternatives I can think of at the moment are `LockUnavailableException` or `LockFailedException`.

camlow325:
> If we end up creating a new Exception class for this, I think it would be best to name it something which does not include the word "pill". As file sync, I don't care at all that the pool uses poison pills - that's an implementation detail, so leaking the word "pill" into the API feels wrong. The alternatives I can think of at the moment are LockUnavailableException or LockFailedException.

Fair enough.  Maybe it's premature to create a new exception class for this anyway.  We could just throw `InterruptedException` in the case that a poison pill has "interrupted" the lock attempt.  If we did that, we'd still uphold the current signature of the `lock` method - since it claims to be able to throw an `InterruptedException`.  We could see what the behavior for shutdown would look like with that implementation and see if we think a more specific exception class is warranted.

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
jpinsonault: > Future Us won't have to look at this and wonder "what happens if it's a different pill?"
Or even a different instance. Not sure how that'd happen in practice but I've read the code thought about that in the past. 

That seems like a good approach

>  I think having .insertPill calls beyond the first essentially be a no-op would be okay and maybe give us some flexibility for later.
Agreed with this as well 👍

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
jpinsonault: Cool, sounds good to me

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
jpinsonault: TIL what `volatile` does

calling it `pill` sounds good to me

test/integration/puppetlabs/services/jruby_pool_manager/jruby_pool_int_test.clj
jpinsonault: Looks like this is an outdated diff, that line has changed

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
jpinsonault: Sorry forgot to reply to this. I'm not sure if it's desirable in the extreme, but I don't know if it's ever come up before as a problem before, since the behavior is the same as before these changes, just using the agent to queue instead of the pool lock

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
camlow325: Since the `allRegisteredInQueue` Condition variable is basically only used in the context of signalling a waiting locker that it can proceed, maybe renaming this to something like `lockAvailable` or something like that would be better, or at least shorter anyway.

jpinsonault:
I'm fine with going with an `InterruptedException` for now

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
jpinsonault: I like `lockAvailable`, easier to understand

jpinsonault:
Pushed some commits that address the latest round of discussion. I think I've caught everything, but it's getting a little difficult to search through the feedback at this point, so speak up if I missed something

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -139,8 +145,11 @@
   [pool-context :- jruby-schemas/PoolContext
    pool-state :- jruby-schemas/PoolState
    refill? :- schema/Bool]
-  (log/info "Draining JRuby pool.")
-  (let [old-instances (collect-all-jrubies pool-state)]
+  (if refill?
+    (log/info "Draining and refilling JRuby pool.")
+    (log/info "Draining JRuby pool."))
+  (let [shutdown-on-error (get-shutdown-on-error-fn pool-context)
+        old-instances (shutdown-on-error #(collect-all-jrubies pool-state))]
```
camlow325: Looking at the current implementation of `collect-all-jrubies`, having a `shutdown-on-error` wrapper here seems okay for now since the only exceptions that would appear to be possible to bubble up here would be for cases where a borrow failed (and a poison pill inserted into the pool).  May need to make the shutdown more surgical in the future, e.g., if/when we implement the lock timeout that we've been talking about for [SERVER-1704](https://tickets.puppetlabs.com/browse/SERVER-1704).

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
```diff
@@ -202,31 +203,50 @@ public E borrowItemWithTimeout(long timeout, TimeUnit unit) throws
...
+     * inserted
+     */
     @Override
     public void releaseItem(E e) {
         final ReentrantLock lock = this.queueLock;
         lock.lock();
         try {
-            // Do nothing if the item being returned is the poisonPill
-            if (e != this.poisonPill){
+            if (!isRegistered(e) && e != this.pill){
```
camlow325: Could just do the `!isRegistered(e)` check and throw an exception inside of the `e != this.pill` check around the `addFirst(e)` call below.  That way, you'd only need to do the `e != this.pill` once.

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
```diff
@@ -399,8 +427,8 @@ private void signalIfAllRegisteredInQueue() {
         // the pool lock but is waiting for all registered elements to be
         // returned to the queue.
         if (registeredElements.size() == liveQueue.size() ||
-                poisonPill != null) {
-            allRegisteredInQueue.signal();
+                pill != null) {
+            lockAvailable.signal();
```
camlow325: Could also change the name of this method to something more inclusive, like `signalIfAllRegisteredOrPillInQueue` or `signalIfLockCanProceed`.  I'm good with whatever seems most reasonable to you here.

test/unit/puppetlabs/jruby_utils/lockable_pool_test.clj
```diff
@@ -591,17 +591,81 @@
...
       (.insertPill pool pill)
       ; borrow finishes and gives us back the pill
-      (let [promise-result (timed-deref borrow-promise)]
-        (is (identical? promise-result pill))))))
+      (let [future-result (timed-deref blocked-borrow)]
+        (is (identical? future-result pill)))))
+
+  (testing "second insert doesn't change the pill"
+    (let [pool (create-populated-pool 1)
+          first-pill (str "pill clinton")
```
camlow325: That's mr. pill to you.

test/unit/puppetlabs/jruby_utils/lockable_pool_test.clj
```diff
@@ -591,17 +591,81 @@
...
+          second-pill (str "pillary clinton")]
+      (.insertPill pool first-pill)
+      (is (identical? first-pill (.borrowItem pool)))
+
+      (.insertPill pool second-pill)
+      ; Should still be equal to the first pill
+      (is (identical? first-pill (.borrowItem pool)))
+      (is (not (identical? second-pill (.borrowItem pool)))))))
```
camlow325: This assertion might be redundant?  Doesn't seem to hurt anything though.

test/unit/puppetlabs/jruby_utils/lockable_pool_test.clj
jpinsonault: twirling towards freedom

test/unit/puppetlabs/jruby_utils/lockable_pool_test.clj
```diff
@@ -591,17 +591,81 @@
...
+          pill "pillful ignorance"
+          instance (.borrowItem pool)]
+      ; Exceptions thrown from the future will be returned as InterruptedException,
+      ; so we can't use thrown-with-msg?. We'll catch it, return it instead of
+      ; throwing it, and inspect it manually below
+      (let [blocked-lock-future (future (try (.lock pool)
+                                             (catch InterruptedException e
+                                               e)))]
+        ; Give future time to start
+        (Thread/sleep 500)
```
camlow325: To avoid the sleep here (and make this a bit less racy), I think you could wait on an `.isLocked()` call for the pool to return true before continuing.  That should happen once the pool lock has been reserved but before the wait on the `lockAvailable` condition, so seems like it would still cover what we want to test here.

jpinsonault:
Pushed a commit addressing the feedback

camlow325:
This all looks good to me so far.  @jpinsonault, I know you had one other bug you were trying to chase down around an `IllegalStateException` being thrown during a shutdown flush.  Would you prefer to hold off on fixing that first before @KevinCorcoran does another pass through this or do you think it would make sense for him to go ahead and look again (assuming he has time to do so).

jpinsonault:
I think it'd be worth looking at (if he has time), since a lot has changed. But also fine to hold off. Either way, I don't think it'd be a waste of time

jpinsonault:
@camlow325  I think I figured out the issue with exception during shutdowns. It seems to happen more often in tests that also flush the pool for whatever reason. I think that, since the queueing happens through the pool lock and not the agent, when 2 flushes happen quickly, the second flush will check if the pool is initialized before the pool has had a chance to refill. It's ok to do that since the lock won't be granted until it's full anyways, but the check is happening before that. 

Re: our discussion yesterday I don't think the check serves any purpose anymore. Seems like there are two situations here:
1. If a flush comes in before the pool has finished refilling after a previous flush, the second flush will just block until the pool is refilled. 
2. If the first flush is due to a shutdown/error, the pool will not be refilled, and a pill will be inserted, causing the second flush's call to `.lock` to throw

So I'd say we should just remove the check.

jpinsonault:
I've removed the check.

I'm done integrating this into puppet-server. After removing that unneeded check, the only problems were with tests concerned with race conditions that aren't possible under the new single pool system. 

I can throw up a PR for that, but I'm not sure if we should merge this until we do some gatling testing though? What do you think @camlow325

camlow325:
@jpinsonault As we'd discussed IRL, I'm okay with removing the `pool-initialized?` check.

I would still like to do at least some gatling testing of this rolled into puppetserver before we merge this one.  Would be cool if we could do that while @rlinehan is still reviewing it.

If you want to put up your corresponding puppetserver PR in the meantime, I could review that too.  Not sure if there's more to it than just removing the couple of tests that are no longer valid with the new jruby-utils implementation.  Seems like there was some work you were going to try to rework the timing a bit so that we could still validate the 503 / shutdown pill behavior.  Not sure if you figured out what is possible to do for that bit?

jpinsonault:
I'll see about the gatling tests then

Here's the PS PR https://github.com/puppetlabs/puppetserver/pull/1313

Didn't figure out a way to pause the shutdown and make sure a 503 is returned, but we can still test that once the shutdown has occurred, a 503 is returned. Technically any requests that come in during a shutdown will block until the shutdown is complete, and then return a 503. So it kind of amounts to the same thing

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
```diff
@@ -221,25 +230,18 @@
   return-to-pool
   "Return a borrowed pool instance to its free pool."
   [instance :- jruby-schemas/JRubyInstanceOrPill]
-  (if (jruby-schemas/jruby-instance? instance)
```
camlow325: I think this `if` statement still needs to be here.  When running the `test-503-when-jruby-is-first-to-shutdown` test from puppetserver, I started hitting schema errors from the `get-instance-state-container` function below during shutdown when a `return-to-pool` call was passed a shutdown poison pill.  If the instance passed in is a shutdown poison pill, I don't think we need to return it to the pool anymore.  But we still should avoid trying to manipulate a `:borrow-count` for it, something which only applies for a real jruby-instance.

The lack of a test failure for this also seems to indicate that maybe we're missing a test here.

jpinsonault: 100% agreement. I already made the changes, but it looks like I forgot to commit them. That's a good point, I only found that this function had a problem once I ran the PS tests on it. I'll add a test in jruby-utils that exercises it

jpinsonault:
Pushed a couple commits for making sure returning poison pills doesn't cause problems

test/unit/puppetlabs/services/jruby_pool_manager/jruby_pool_test.clj
```diff
@@ -325,3 +326,22 @@
...
+     app
+     jruby-testutils/default-services
+     {}
+     (let [config (jruby-test-config 2)
+           pool-manager-service (tk-app/get-service app :PoolManagerService)
+           pool-context (pool-manager-protocol/create-pool pool-manager-service config)
+           pool (jruby-core/get-pool pool-context)
+           pill (ShutdownPoisonPill. pool)]
+       ; Returning a pill should be a noop
+       (jruby-core/return-to-pool pill :test [])))))
```
camlow325: I guess you could put an `(is (true? true)` kind of assertion after the `return-to-pool` call so it's easy to see in the repl that the test actually ran?  Seems like we've done that in a number of other places where we're just trying to test to see that we can get past a certain point without throwing an exception.

camlow325:
Last couple of commits look good to me - just one nit on the new `return-to-pool` test.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
rlinehan: Isn't there an `insert-poison-pool` in jruby-internal now that we can use?

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -86,87 +87,102 @@
...
-  "Replace the current pool with a new pool and drain the old pool,
-  optionally refilling the new pool with fresh jrubies."
+(schema/defn collect-all-jrubies :- [JRubyInstance]
+  "Locks the pool and borrows all the instances"
+  [pool-state :- jruby-schemas/PoolState]
+  (let [pool-size (:size pool-state)
+        pool (:pool pool-state)
+        borrow-fn (fn [_] (jruby-internal/borrow-from-pool!*
+                           jruby-internal/borrow-without-timeout-fn
+                           pool))]
```
rlinehan: I think that instead of passing in the `pool-state` here, you could pass in the `pool-context` and then use `jruby-internal/get-pool-size` and `jruby-internal/borrow-from-pool`, both of which take the `pool-context`. That way you don't have to basically recreate `borrow-from-pool` here.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
jpinsonault: yeah, good catch

test/integration/puppetlabs/services/jruby_pool_manager/jruby_pool_int_test.clj
```diff
@@ -142,6 +116,7 @@
                         [])
                        (inc loop-count)))))))
 
+
```
rlinehan: Extra line?

test/integration/puppetlabs/services/jruby_pool_manager/jruby_pool_int_test.clj
```diff
@@ -173,63 +148,36 @@
      (let [config (jruby-testutils/jruby-config {:max-active-instances 4
                                                  :borrow-timeout test-borrow-timeout})
            pool-manager-service (tk-app/get-service app :PoolManagerService)
-           pool-context (pool-manager-protocol/create-pool pool-manager-service config)]
+           pool-context (pool-manager-protocol/create-pool pool-manager-service config)
+           pool-state (jruby-internal/get-pool-state pool-context)
+           pool (:pool pool-state)]
```
rlinehan: Doesn't seem like `pool-state` is being used, so I think you could just use `jruby-internal/get-pool`.

test/unit/puppetlabs/services/jruby_pool_manager/jruby_pool_test.clj
rlinehan: Agree with @camlow325 that this would be a good thing to do.

test/unit/puppetlabs/services/jruby_pool_manager/jruby_testutils.clj
```diff
@@ -86,3 +87,34 @@
...
+   (wait-for-predicate f 20 500))
+  ([f :- IFn
+    max-iterations :- schema/Int
+    sleep-time :- schema/Int]
+   (loop [count 0]
+     (cond (f)
+           true
+
+           (>= count max-iterations)
+           false
```
rlinehan: Should it log something in this case to help test debugging?

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -86,87 +87,102 @@
    pool :- jruby-schemas/pool-queue-type]
   (= expected-pool-size (count (.getRegisteredElements pool))))
 
-(schema/defn ^:always-validate
-  swap-and-drain-pool!
-  "Replace the current pool with a new pool and drain the old pool,
-  optionally refilling the new pool with fresh jrubies."
+(schema/defn collect-all-jrubies :- [JRubyInstance]
```
jpinsonault: I like `borrow-all-jrubies`

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -86,87 +87,102 @@
...
+   old-instances :- [JRubyInstance]
    refill? :- schema/Bool]
-  (let [{:keys [config]} pool-context
-        pool-state-atom (jruby-internal/get-pool-state-container pool-context)
-        new-pool (:pool new-pool-state)
-        old-pool (:pool old-pool-state)
-        old-pool-size (:size old-pool-state)
+  (let [pool (:pool pool-state)
+        pool-size (:size pool-state)
+        new-instance-ids (map inc (range pool-size))
```
jpinsonault: That's a really good question. I can't find anything to indicate that the `id` is even used anywhere

camlow325: Hmm, good point.  I *think* the ids were only used just for troubleshooting / diagnostics purposes.  Like you could see "Flushing instance [id]" in the log when an instance was flushed.  So I don't know that the current set of changes on this PR would functionally break anything other than the usability of the log messages, which would still be a shame.

I remember debating with Chris at one point about whether we should just have an atomic id counter in the pool that we increment by one and hand off to the instance each time we create a new one.  Maybe this would be a good time to do that now?  I'd be okay with doing that as a follow-on ticket to this one, though, to avoid holding this one up further - but maybe one we should do before we do the next jruby-utils release?

@rlinehan, @jpinsonault - WDYT?

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
```diff
@@ -232,14 +244,8 @@
...
-          (try
-            (flush-instance-fn pool instance)
-            (finally
-              (.releaseItem pool instance false))))
-        (.releaseItem pool instance)))
-    ;; if we get here, it was from a borrow and we got a Retry, so we just
-    ;; return it to the pool.
-    (.releaseItem (:pool instance) instance)))
+          (flush-instance-fn instance))
+        (.releaseItem pool instance)))))
```
jpinsonault: So before, that `(finally (.releaseItem pool instance false))` block was a noop. Passing in `false` to `.releaseItem` became a noop at some point in the past I think. @camlow325 might know why. But we decided to remove that overload of `.releaseItem`

The `flush-instance-fn` (used to and still does) unregisters the old instance, creates the new instance, and adds it to the pool

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
jpinsonault: @camlow325 ahh, ok. Yeah we should preserve that then. There's a `next-instance-id` fn that guarantees a unique new id based off the old id. We could use that if we don't want to use an atom.

Sounds like a good followup PR

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -86,87 +87,102 @@
...
-          new-pool (:pool new-pool-state)
-          old-pool-state (jruby-internal/get-pool-state pool-context)
-          old-pool (:pool old-pool-state)
-          old-pool-size (:size old-pool-state)]
-      (when-not (pool-initialized? old-pool-size old-pool)
-        (throw (IllegalStateException. "Attempting to flush a pool that does not appear to have successfully initialized. Aborting.")))
-      (.insertPill new-pool (ShutdownPoisonPill. new-pool))
-      (swap-and-drain-pool! pool-context old-pool-state new-pool-state false))
-    (finally
-      (deliver on-complete true))))
```
camlow325: Hmm, good question.  I don't think we need to have the ability to block until the pool instances have been totally destroyed / refilled in the case of a pool flush.  In the case of a shutdown, though, I do think it makes sense to have the ability to block until all of the old instances have been destroyed.  Without the promise in this function anymore, I'm not sure that we still have this capability.  @jpinsonault - is there something we're missing here?

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
camlow325: I think `next-instance-id`'s current implementation might not work for this because it just tried to determine a new id based on the current id + the size of the current pool.  That, I think, generally worked for the old pool implementation because we only had individual instance flushes and not full pool flushes to deal with for the same pool - since the full pool ones always resulted in the creation of a new pool.  I think it would be more useful from the perspective of the log messages overall to just assign an id that is more guaranteed to be unique per pool instance over the life of a puppetserver service run.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
camlow325: Yeah, what @jpinsonault said sounds right.  We had the `.releaseItem` overload in the past because the underlying Java concurrency construct that we we were using for locking needed to be decremented each time we were done with borrowing an instance (to avoid a deadlock later on) - even though we didn't end up actually returning the instance back to the pool because we wanted to dispose of it.  That became irrelevant in the later JRubyPool implementation because we didn't have a need to decrement a lock counter when disposing of an item anymore.  We only kept the no-op overload at the time for backward compatibility in the API.  I can't exactly remember why now, but I think I suggested to @jpinsonault that he go ahead and remove the unnecessary overload as part of this PR because it seemed to simplify some other part of the rework he was doing.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -182,27 +198,19 @@
   send-prime-pool! :- jruby-schemas/JRubyPoolAgent
   "Sends a request to the agent to prime the pool using the given pool context."
   [pool-context :- jruby-schemas/PoolContext]
-  (let [pool-agent (get-pool-agent pool-context)]
-    (send-agent pool-agent #(prime-pool! pool-context))))
+  (let [modify-instance-agent (get-modify-instance-agent pool-context)]
+    (send-agent modify-instance-agent #(prime-pool! pool-context))))
 
 (schema/defn ^:always-validate
   send-flush-and-repopulate-pool! :- jruby-schemas/JRubyPoolAgent
```
camlow325: Sounds reasonable to me since I don't think we use this outside of jruby-utils.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
jpinsonault: @camlow325 yeah I think you're right. It would be nice to have a way to tell when the flush has completed. Right now when this function returns you only know that all the instances have been borrowed. Adding the promise back in seems like a good idea. 

The `flush-and-repopulate-pool!` fn can just ignore the promise since it doesn't need to block until the instances have been totally destroyed and recreated

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
```diff
@@ -229,11 +229,7 @@
   "Flush all the current JRubyInstances so that the pool can be shutdown
   without any instances being active."
   [pool-context]
-  (let [on-complete (promise)]
-    (log/debug "Beginning flush of JRuby pools for shutdown")
-    (jruby-agents/send-flush-pool-for-shutdown! pool-context on-complete)
-    @on-complete
-    (log/debug "Finished flush of JRuby pools for shutdown")))
+  (jruby-agents/flush-pool-for-shutdown! pool-context))
```
camlow325: Yeah, I agree with that (mentioned in a comment on the `drain-and-refill-pool!` function) - good catch.

jpinsonault:
@camlow325 @rlinehan I've pushed some commits to address @rlinehan's feedback

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
```diff
@@ -293,8 +331,12 @@ public void lock() throws InterruptedException {
                 }
             }
             try {
-                while (registeredElements.size() != liveQueue.size()) {
-                    allRegisteredInQueue.await();
+                //
```
camlow325: Nit, seems like this empty comment could go away - or you could replace it with something which says we want to hold here until the pool has been filled to capacity.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -202,29 +201,26 @@
...
-  ;;    control until it has borrowed the correct number of instances.
-  ;; 2. While that is happening, an individual instance reaches the 'max-borrows' value.  This instance will never
-  ;;    be returned to the pool; it is handled by sending a function to an agent, which will flush the individual
-  ;;    instance, create a replacement one, and return that to the pool.
-  ;;
-  ;; If we use the same agent for both of these operations, then step 2 will never begin until step 1 completes, and
-  ;; step 1 will never complete because the `max-borrows` instance will never be returned to the pool.
-  ;;
-  ;; Using a separate agent for the 'max-borrows' instance flush alleviates this issue.
+  ;; We use an agent to syncronize jruby creation and destruction to migitage
```
rlinehan: Nitpick: s/migitage/mitigate/

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -202,29 +201,26 @@
     (send-agent modify-instance-agent #(prime-pool! pool-context))))
 
 (schema/defn ^:always-validate
-  send-flush-and-repopulate-pool! :- jruby-schemas/JRubyPoolAgent
-  "Sends requests to the agent to flush the existing pool and create a new one."
+flush-and-repopulate-pool!
```
rlinehan: Nitpick: I think the alignment here is off?

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -202,29 +201,26 @@
...
+  "Flush of the current JRuby pool. Blocks until all the instances have
+  been borrowed from the pool, and before the pool has been refilled"
```
rlinehan: I might just not be awake enough yet, but I'm having a hard time parsing this.

rlinehan: The "and before the pool has been refilled" part

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
```diff
@@ -222,18 +222,16 @@
...
   flush-pool-for-shutdown!
   "Flush all the current JRubyInstances so that the pool can be shutdown
   without any instances being active."
   [pool-context]
   (let [on-complete (promise)]
-    (log/debug "Beginning flush of JRuby pools for shutdown")
-    (jruby-agents/send-flush-pool-for-shutdown! pool-context on-complete)
-    @on-complete
-    (log/debug "Finished flush of JRuby pools for shutdown")))
+    (jruby-agents/flush-pool-for-shutdown! pool-context on-complete)
```
camlow325: Do you think we still need an `on-complete` promise at this level?  It seems like we could just create/use the promise behind the `jruby-agents/flush-pool-for-shutdown!` function to hold up the call until the flush has been completed and not need to let that percolate up here.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -62,111 +61,115 @@
...
-  (let [{:keys [config]} pool-context
-        new-pool-state (jruby-internal/create-pool-from-config config)
-        old-pool (jruby-internal/get-pool-state pool-context)]
-    (swap-and-drain-pool! pool-context old-pool new-pool-state true)))
+  (log/debug "Beginning flush of JRuby pools for shutdown")
+  (let [pool-state (jruby-internal/get-pool-state pool-context)
+        pool (:pool pool-state)]
+    (drain-and-refill-pool! pool-context false on-complete)
+    (jruby-internal/insert-shutdown-poison-pill pool))
+  (log/debug "Finished flush of JRuby pools for shutdown")
```
camlow325: Seems like the "Finished flush" log statement should be after the `@on-complete` one?

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -62,111 +61,115 @@
...
-                  e)))))
-    ;; Add a "RetryPoisonPill" to the pool in case something else is in the
-    ;; process of borrowing from the old pool.
-    (.insertPill old-pool (RetryPoisonPill. old-pool))))
+                  "There was a problem creating a JRubyInstance for the pool."
+                  e))))))
+  (if refill?
+    (log/info "Finished draining and refilling pool.")
+    (log/info "Finished draining pool."))
+  (deliver on-complete true))
```
camlow325: We should make sure that the `on-complete` promise is delivered even if an exception occurs during the cleanup/creation loop above, not just if we got all the way through the loop successfully.  Since the only function which calls `cleanup-and-refill-pool` appears to be `drain-and-refill-pool!`, I wonder if it would be better to set the promise from within the call to `cleanup-and-refill-pool` instead, and not bother passing the promise in here at all.

So, maybe in `drain-and-refill-pool!`, we could instead have:

```clj
  (send-agent (get-modify-instance-agent pool-context)
                 #(try
                      (cleanup-and-refill-pool pool-context old-instances refill?)
                      (finally
                           (deliver on-complete true))
```

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
jpinsonault: good idea

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
jpinsonault: Ah, yeah it's weird. I'm trying to say that it will return once the pool instances have been borrowed, and that it won't wait for the instances to be destroyed and recreated.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
jpinsonault: yeah good call

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
jpinsonault: Yeah I'm not sure why I decided to have it up at this level in retrospect, I'll change it

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
jpinsonault: wow that's almost impressive

jpinsonault:
Pushed some commits to address the latest round of feedback

rlinehan:
👍 💯 :shipit:

PR #35, opened 2016-12-16
(SERVER-1475) Add an JRuby slf4j Logger adapter

camlow325:
This commit adds a class, Slf4jLogger, which implements the JRuby Logger
interface, forwarding any logging calls which are received on to an
slf4j logger.  This class doesn't actually hook up the logger to created
JRuby instances since JRuby classes typically get the object that they
use for logging during static initialization, which would most likely
complete before any attempt to set a custom logger with JRuby could be
done.  Use of this Logger would typically require that the consuming
process set the jruby.logger.class Java system property to the namespace
of the new class on the command line, e.g.,
-Djruby.logger.class=com.puppetlabs.jruby_utils.jruby.Slf4jLogger.

### DIFF
--- a/test/unit/puppetlabs/jruby_utils/slj4j_logger_test.clj
+++ b/test/unit/puppetlabs/jruby_utils/slj4j_logger_test.clj
@@ -1,13 +1,13 @@
 (ns puppetlabs.jruby-utils.slj4j-logger-test
   (:require [clojure.test :refer :all]
             [puppetlabs.trapperkeeper.testutils.logging :as logutils])
-  (:import (com.puppetlabs.jruby_utils.jruby Slf4jLogger)))
+  (:import (org.jruby.util.log LoggerFactory)))
 
 (deftest slf4j-logger-test
   (let [actual-logger-name "my-test-logger"
         exception-message "exceptionally bad news"
         expected-logger-name (str "jruby." actual-logger-name)
-        logger (Slf4jLogger. actual-logger-name)
+        logger (LoggerFactory/getLogger actual-logger-name)
         actual-log-event (fn [event]
                            (assoc event :exception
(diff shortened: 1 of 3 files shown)


test/unit/puppetlabs/jruby_utils/slj4j_logger_test.clj
```diff
@@ -0,0 +1,93 @@
...
+                              :level level
+                              :exception exception
+                              :logger expected-logger-name})]
+    (testing "name stored in logger"
+      (is (= expected-logger-name (.getName logger))))
+
+    (testing "warn with a string and objects"
+      (logutils/with-test-logging
+       (.warn logger "a {} {} warning" (into-array Object ["strongly" "worded"]))
+       (is (logged? "a {} {} warning" :warn))))
```
camlow325: I would have liked to have some test assertions here confirm that the `Object` array from above, from the `(into-array Object ["strongly" "worded"])` form, is passed on through to the logging framework.  Unfortunately, though, the logic behind the `logged?` function only grabs the message and not any associated arguments.  I did at least do some manual validation of logged messages to see that the contents of the`Object` array are being subbed into the actual message which gets logged, for example:

> 2016-12-16 06:48:09,545 WARN  [nREPL-worker-4] [j.my-test-logger] a strongly worded warning

I could look into plumbing the extra parameter args through `logged?`.  Not sure how valuable this would be to do for the effort this might involve, though.  Interested in others opinions on that.

cprice404: Doesn't seem worth it

rlinehan: Agree Chris that that doesn't seem worth it.

PR #40, opened 2017-02-10
Server 1704 introduce pool lock timeout

jpinsonault:
This PR modifies the pool flush code to use a lock timeout.

It also adds a new config setting to specify the timeout.

---

This PR goes along with https://github.com/puppetlabs/puppetserver/pull/1330

### DIFF
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
@@ -4,9 +4,11 @@
             [clojure.tools.logging :as log]
             [puppetlabs.kitchensink.core :as ks]
             [puppetlabs.services.jruby-pool-manager.jruby-schemas :as jruby-schemas]
-            [puppetlabs.i18n.core :as i18n])
+            [puppetlabs.i18n.core :as i18n]
+            [slingshot.slingshot :as sling])
   (:import (clojure.lang IFn IDeref)
-           (puppetlabs.services.jruby_pool_manager.jruby_schemas PoisonPill JRubyInstance)))
+           (puppetlabs.services.jruby_pool_manager.jruby_schemas PoisonPill JRubyInstance)
+           (java.util.concurrent TimeUnit TimeoutException)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Private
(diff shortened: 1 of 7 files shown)


jpinsonault:
`jruby_agents.clj` is easier to read using the split view

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -86,24 +88,55 @@
...
+    ; shutdown-on-error will either return the jrubies as expected,
+    ; or if there was an error, it will shutdown the server and return a promise.
+    ; We want to rethrow whatever exception it encountered so that it can bubble
+    ; up to whatever code requested this flush, so borrow-all-jrubies* will put
+    ; that exception into the borrow-exception atom
+    (let [jrubies (shutdown-on-error #(borrow-all-jrubies* pool-context borrow-exception))]
+      (if-let [exception @borrow-exception]
+        (throw exception)
+        jrubies))))
+
```
jpinsonault: This `borrow-all-jrubies`/`borrow-all-jrubies*` stuff is where all the jank is, and where feedback is needed most probably

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -86,24 +88,55 @@
...
+  (let [pool (jruby-internal/get-pool pool-context)
+        flush-timeout (jruby-internal/get-flush-timeout pool-context)
+        shutdown-on-error (get-shutdown-on-error-fn pool-context)
+        borrow-exception (atom nil)]
+    ; If lock fails, abort and throw an exception
+    (try
+      (.lockWithTimeout pool flush-timeout TimeUnit/MILLISECONDS)
+      (catch TimeoutException e
+        (sling/throw+ {:kind ::jruby-lock-timeout
+                       :msg (.getMessage e)})))
```
jpinsonault: This expands to `:puppetlabs.services.jruby-pool-manager.impl.jruby-agents/jruby-lock-timeout` which puppetserver or any other consumer has to reference. I'm not sure what the best way to handle this is. Maybe just explicitly throw a `:puppetlabs.services.jruby-pool-manager.jruby-core/jruby-lock-timeout` or something?

jpinsonault:
Ping @camlow325 @rlinehan for review when you get a chance

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
camlow325: I think this approach is fine.  Since we need to have the exception be handled by the `shutdown-on-error` macro - in order for it to trigger the shutdown - but we don't have the ability to control what `shutdown-on-error` returns to the caller in that case, I think this might be the best we can do here without making changes in Trapperkeeper.

jpinsonault:
Pushed some commits responding to feedback

camlow325:
This all looks fine to me.  @rlinehan - did you have any feedback on this?

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
```diff
@@ -491,8 +491,7 @@ private void signalIfLockCanProceed() {
         // is active at a time - a caller of lock() that has just acquired
         // the pool lock but is waiting for all registered elements to be
         // returned to the queue.
-        if (registeredElements.size() == liveQueue.size() ||
-                pill != null) {
+        if (this.maxSize == liveQueue.size() || pill != null) {
```
rlinehan: This change scares me a bit, I think because I haven't quite been able to think through why we were using `registeredElements` before rather than `maxSize` and when we register/unregister instances and ahhh my brain.

Um, anyway, other than me being just possibly irrationally nervous about this, I think the comment is now out of date - we're not waiting for all registered elements to be returned to the queue, we're waiting for the queue to be the right size.

camlow325:
> Sort of related: I left a bunch of comments on CODEMGMT-1013.

@rlinehan Thanks much for doing that.  In [that ticket comment](https://tickets.puppetlabs.com/browse/CODEMGMT-1013?focusedCommentId=410106), you obviously pointed out several things I hadn't been thinking about when I wrote up the description for SERVER-1704.

I had originally been thinking of SERVER-1704 as being the ticket where we'd do all of the OSS Puppet Server and jruby-utils work needed to support the file sync lock timeout use case.  In particular, having a `with-lock-with-timeout` macro in jruby_utils' jruby-core namespace which would call into a new `lock-pool-with-timeout` function would seem to make sense.  That `lock-pool-with-timeout` function could then call into the new `.lockWithTimeout` method that was already added to the JRubyPool class for this PR.

@jpinsonault and @rlinehan - would it make sense to you to go ahead and add those extra helper functions onto this PR or do those a separate ticket?  I'm inclined to just add them onto the PR but wanted to see what you thought.

rlinehan:
@jb ¯\_(ツ)_/¯ I don't really care much either way? This PR's been up for 2 weeks already, so might be good to just close it and open a new one for the other stuff, but really whatever.

src/java/com/puppetlabs/jruby_utils/pool/JRubyPool.java
camlow325: We did made this change in the `lock()` method as part of #34.  I think @jpinsonault was just cleaning up this other reference in `signalIfLockCanProceed()` to do the same.

I originally had had this key off of `registeredElements` rather than `liveQueue` size because I was thinking of a future world in which we might be able to dynamically size the `maxSize` of the pool.  For that case, I had been thinking that if a lock were requested that the unregistering of an element from the pool should, assuming all other still registered instances are in the pool, allow the lock to be granted/proceed.  For #34, we changed the flush implementation to acquire a lock before borrowing all instances but, for performance, release the lock before filling the pool back up.  For the case where two flush lock requests and one or more separate borrow requests overlap one another, waiting for `registeredElements` to equal `liveQueue` size created a race condition where the second lock request could be granted before the pool was filled back up and, thereby, create a "deadlock" where the second locker would hang forever when trying to borrow an instance out of the pool - because the separate borrow request was signaled that the instance was available to be borrowed rather than the second locker.

I can think of some other ways we could have worked around this problem without giving up the ability to let the pool be dynamically sizable in the future.  We don't actually have a requirement yet for a dynamically sizable pool, though - and may never have one.  This seemed like a much simpler change to head off the "deadlock" problem than other options.

Hope some of that made sense.  @jpinsonault admirably spent a lot of time diagnosing this problem when it first came up and had us both scratching our heads for quite a while before we were able to figure out what was going on.

camlow325: In any event, I agree that the comment in the code should be updated to talk about the max size/capacity of the pool rather than registered elements.

rlinehan: Thanks for the explanation @camlow325! I think I was missing the bit that we were originally using `registeredElements` rather than `maxSize` in the past because we hoped to do dynamic sizing at some point - if that was our only requirement for having that then this change all makes sense to me (and I'm less scared about it :D). 👍

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -86,24 +88,55 @@
     (jruby-internal/create-pool-instance! pool new-id config
                                           (partial send-flush-instance! pool-context))))
 
-(schema/defn borrow-all-jrubies :- [JRubyInstance]
-  "Locks the pool and borrows all the instances"
-  [pool-context :- jruby-schemas/PoolContext]
+(schema/defn borrow-all-jrubies*
+  "The core logic for borrow-all-jrubies. Should not be called from borrow-all-jrubies"
```
jpinsonault: Ha, yeah it says the exact opposite of what it should

jpinsonault:
I'm happy with fixing up the last few feedback items, merging this, and opening a new PR for the other changes

jpinsonault:
Pushed commits for the latest feedback

camlow325:
LGTM so I'm going to merge it.  @rlinehan, if you see something else we missed in the latest review, maybe @jpinsonault can pick that up in his follow up PR.

PR #45, opened 2017-04-21
(SERVER-1630) Add support for exercising JRuby 9k

camlow325:
This PR adds support for exercising JRuby 9k in jruby-utils.  The default
jruby-utils dependencies still pull in dependencies from JRuby 1.7.x.  When
an alternate leiningen profile, jruby9k, is exercised when `lein test` is run,
however, jruby9k-related dependencies are pulled in instead of JRuby
1.7.x-related ones.

This PR also includes some code specific to how jruby-utils is exercised
when running with JRuby 9k, including:

* Emit a warning if the `compat-version` setting is set.  JRuby 9k has a
   hardcoded `compat-version` set internally and, so, it cannot be
   overridden at run-time.

* Code for cleaning up temporary jars and jar files which are currently
   being leaked by JRuby when a ScriptingContainer is terminated.  See
   SERVER-1777 for more details on pursuing a longer-term fix for this
   issue in JRuby itself.

### DIFF
--- a/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
@@ -26,7 +26,9 @@
 
 (def default-jruby-compat-version
   "Default value for JRuby's 'CompatVersion' setting."
-  "1.9")
+  (if (contains? jruby-schemas/supported-jruby-compat-versions "1.9")
+    "1.9"
+    (first jruby-schemas/supported-jruby-compat-versions)))
 
 (def default-borrow-timeout
   "Default timeout when borrowing instances from the JRuby pool in
@@ -164,39 +166,45 @@
           (integer? compat-version) double
           :true str))
 
-(schema/defn ^:always-validate warn-if-compat-version-setting-not-supported
-  [parsed-compat-version :- schema/Str]
-  (let [const-compat-version (jruby-internal/get-compat-version
-                              parsed-compat-version)
-        instance-config (RubyInstanceConfig.)]
-    (.setCompatVersion instance-config const-compat-version)
-    (when-not (= const-compat-version (.getCompatVersion instance-config))
-      (log/warnf
-       "%s %s %s"
-       (i18n/trs "compat-version is set to `{0}`, which is not a supported version."
-                 parsed-compat-version)
-       (i18n/trs "Version `{0}` will be used instead." Constants/RUBY_VERSION)
-       (i18n/trs "Setting 'compat-version' is deprecated and will be removed in a future release.")))))
+(schema/defn ^:always-validate get-compat-version-for-jruby-config
+  [compat-version-from-user-config :- schema/Any
+   default-compat-version :- schema/Str
+   supported-compat-versions :- #{schema/Str}]
+  (let [parsed-compat-version (parse-compat-version
+                               (or compat-version-from-user-config
+                                   default-compat-version))]
+    (when (and compat-version-from-user-config
+               (= 1 (count supported-compat-versions)))
+      (log/warn
+       (i18n/trs "Setting 'compat-version' for JRuby 9k is deprecated and will be removed in a future release.")))
+    (if (contains? supported-compat-versions parsed-compat-version)
+      parsed-compat-version
+      (do
+        (log/warnf
+         "%s %s"
+         (i18n/trs "compat-version is set to `{0}`, which is not a supported version."
+                   parsed-compat-version)
+         (i18n/trs "Version `{0}` will be used instead." default-compat-version))
+        default-compat-version))))
 
 (schema/defn ^:always-validate
   initialize-config :- jruby-schemas/JRubyConfig
   "Initialize keys with default settings if they are not given a value.
   The config is validated after these defaults are set."
   [{:keys [compat-version] :as config} :- {schema/Keyword schema/Any}]
-  (let [parsed-compat-version (parse-compat-version
-                               (or compat-version default-jruby-compat-version))]
-    (when compat-version
-      (warn-if-compat-version-setting-not-supported parsed-compat-version))
-    (-> config
-        (update-in [:compile-mode] #(keyword (or % default-jruby-compile-mode)))
-        (assoc :compat-version parsed-compat-version)
-        (update-in [:borrow-timeout] #(or % default-borrow-timeout))
-        (update-in [:flush-timeout] #(or % default-flush-timeout))
-        (update-in [:max-active-instances] #(or % (default-pool-size (ks/num-cpus))))
-        (update-in [:max-borrows-per-instance] #(or % 0))
-        (update-in [:environment-vars] #(or % {}))
-        (update-in [:lifecycle] initialize-lifecycle-fns)
-        jruby-internal/initialize-gem-path)))
+  (-> config
+      (update-in [:compile-mode] #(keyword (or % default-jruby-compile-mode)))
+      (assoc :compat-version (get-compat-version-for-jruby-config
+                              compat-version
+                              default-jruby-compat-version
+                              jruby-schemas/supported-jruby-compat-versions))
+      (update-in [:borrow-timeout] #(or % default-borrow-timeout))
+      (update-in [:flush-timeout] #(or % default-flush-timeout))
+      (update-in [:max-active-instances] #(or % (default-pool-size (ks/num-cpus))))
+      (update-in [:max-borrows-per-instance] #(or % 0))
+      (update-in [:environment-vars] #(or % {}))
+      (update-in [:lifecycle] initialize-lifecycle-fns)
+      jruby-internal/initialize-gem-path))
 
 (schema/defn register-event-handler
   "Register the callback function by adding it to the event callbacks atom on the pool context."
--- a/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
@@ -13,7 +13,9 @@
             [puppetlabs.i18n.core :as i18n])
   (:import (puppetlabs.services.jruby_pool_manager.jruby_schemas JRubyInstance)
            (clojure.lang IFn)
-           (java.util.concurrent TimeUnit)))
+           (java.util.concurrent TimeUnit)
+           (org.jruby RubyInstanceConfig)
+           (org.jruby.runtime Constants)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Constants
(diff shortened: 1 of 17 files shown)


camlow325:
There is some work we're doing in ezbake to enable Puppet Server to consume JRuby 1.7 vs. 9k jars, see [SERVER-1772](https://tickets.puppetlabs.com/browse/SERVER-1772).  We may want to hold off on merging this until that work comes together but I wanted to get this up for early review so others could take a look.

camlow325:
Hmm, this failed Travis for JRuby 9k on JDK 7 only with an apparent hang (in two different locations), but not JDK 8:

```
lein test puppetlabs.services.jruby-pool-manager.jruby-pool-test
Tests failed.
Error encountered performing task 'test' with profile(s): 'base,system,user,provided,dev,jruby9k'
Tests failed.
```

```
lein test puppetlabs.services.jruby-pool-manager.jruby-pool-int-test
Tests failed.
Error encountered performing task 'test' with profile(s): 'base,system,user,provided,dev,jruby9k'
Tests failed.
```

Ultimately, we may not ever end up using JRuby 9k on JDK 7 but I'd like to avoid branching jruby-utils for JDK 7 vs. 8 for now if we can and we should keep JDK 7 in the Travis matrix for pre-Puppet 5 support.  I haven't been able to reproduce this failure locally yet on JDK 7 but I do wonder if this is due to excessive memory usage / GC causing the tests to time out.  Will keep poking at it.

camlow325:
Although I'll try just re-running on Travis first to see if I can reproduce.

camlow325:
I'm seeing pretty frequent, but intermittent failures both on JDK 7 and 8 with this PR.  I'm going to put some debug-related commits up on the PR to see how Travis responds.

camlow325:
ok, I think I've finally figured out how to make these tests pass consistently in Travis.  I believe the root cause of the periodic failures was that the test process was exceeding memory limits - although it's hard to tell for sure from looking at the output Travis gives you.  When I bumped the maximum heap up to 2G, in 737888f, the tests started passing much more frequently in Travis across all cells.

Even with the maximum heap size bump, though, I was able to reproduce occasional failures both in Travis and locally when running the tests for JRuby 9k - mostly with exceptions related to failures to load jar files.  The existing unit test suite did not, for the most part, have any logic to call `.terminate()` on the various `ScriptingContainer` objects that they were creating.  This would result in container termination being deferred out to whenever the container's finalizer() method might be called.  I believe the jar file failures could be due to a number of conditions - e.g., file descriptor or native memory exhaustion.

When I changed each of the tests to ensure that `.terminate()` was called on every `ScriptingContainer` which had been created, in b08a9b0, I was no longer able to recreate a test failure.  I've re-run the job in Travis around 20 times or so - and 50 or so times locally - without seeing anymore failures.

src/java/com/puppetlabs/jruby_utils/jruby/InternalScriptingContainer.java
```diff
@@ -10,12 +21,150 @@
...
+            classLoaderTempDirMethod = JRubyClassLoader.class.getDeclaredMethod(
+                    "getTempDir");
+            classLoaderTempDirMethod.setAccessible(true);
+        } catch (ClassNotFoundException|
+                NoSuchFieldException|
+                NoSuchMethodException|
+                IllegalAccessException ex) {
+            jarIndexCacheException = ex;
+        }
+    }
```
jpinsonault: I feel like I'm learning forbidden knowledge reading this

test/integration/puppetlabs/services/jruby_pool_manager/jruby_internal_test.clj
```diff
@@ -27,11 +28,12 @@
                  (jruby-internal/get-compile-mode :foo)))))
 
 (deftest ^:integration settings-plumbed-into-jruby-container
-  (testing "setting plumbed into jruby container for"
+  (testing "settings plumbed into jruby container"
     (let [pool (JRubyPool. 1)
-          config (jruby-testutils/jruby-config
-                  {:compile-mode :jit
-                   :compat-version 2.0})
+          config (logutils/with-test-logging
```
jpinsonault: Is this just a noisy test?

camlow325: Yeah, it's noisy on JRuby 9k because there's some code which emits a deprecation warning when the `:compat-version` is set and JRuby 9k is being used.

jpinsonault:
I'm 👍  on this. The clojure changes look good and as far as I understand the implications of the scripting container additions I'm good with them

rlinehan:
I think it would be really helpful at some point during startup to log what version of JRuby is being used.

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
```diff
@@ -162,21 +164,39 @@
...
+                              parsed-compat-version)
+        instance-config (RubyInstanceConfig.)]
+    (.setCompatVersion instance-config const-compat-version)
+    (when-not (= const-compat-version (.getCompatVersion instance-config))
+      (log/warnf
+       "%s %s %s"
+       (i18n/trs "compat-version is set to `{0}`, which is not a supported version."
+                 parsed-compat-version)
+       (i18n/trs "Version `{0}` will be used instead." Constants/RUBY_VERSION)
+       (i18n/trs "Setting 'compat-version' is deprecated and will be removed in a future release.")))))
```
rlinehan: I wonder if we should mention something about JRuby 9k in here, since (at least as far as I can tell) this message will only be logged when running with JRuby 9k and `compat-version` set?

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
```diff
@@ -162,21 +164,39 @@
...
-      (update-in [:max-active-instances] #(or % (default-pool-size (ks/num-cpus))))
-      (update-in [:max-borrows-per-instance] #(or % 0))
-      (update-in [:environment-vars] #(or % {}))
-      (update-in [:lifecycle] initialize-lifecycle-fns)
-      jruby-internal/initialize-gem-path))
+  [{:keys [compat-version] :as config} :- {schema/Keyword schema/Any}]
+  (let [parsed-compat-version (parse-compat-version
+                               (or compat-version default-jruby-compat-version))]
+    (when compat-version
+      (warn-if-compat-version-setting-not-supported parsed-compat-version))
```
rlinehan: It's a little bit weird that in this case (`compat-version` is set and we log a warning and use JRuby 9k's built in version) we ultimately set `compat-version` below too, especially since this is then logged as part of the JRuby config during startup.

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
rlinehan: Also is there any way we could add a test for this?

src/java/com/puppetlabs/jruby_utils/jruby/InternalScriptingContainer.java
```diff
@@ -10,12 +21,150 @@
...
+            Object jarCache = jarCacheField.get(jarResourceClass);
+            Class jarCacheClass = jarCache.getClass();
+            Field indexCacheField = jarCacheClass.getDeclaredField("indexCache");
+            indexCacheField.setAccessible(true);
+
+            @SuppressWarnings("unchecked") Map<String, Object> indexCache =
+                    (Map<String, Object>) indexCacheField.get(jarCache);
+            jarIndexCache = indexCache;
+
+            Class<?> jarIndexClass = Class.forName("org.jruby.util.JarCache$JarIndex");
```
rlinehan: Pardon my lack of java knowledge... what does `Class<?>` mean here?

rlinehan:
With regards to the last commit... at some point I filed TK-380 to potentially add a `stop` lifecycle function that would clean up jruby instances. What are your thoughts on that?

camlow325:
> With regards to the last commit... at some point I filed TK-380 to potentially add a stop lifecycle function that would clean up jruby instances. What are your thoughts on that?

Thanks, I'd forgotten about that one.  TBH, I'm not sure if I agree with the idea of implementing a `stop` lifecycle method for the `jruby-pool-manager-service` which would cleanup pools.  I think of the `create-pool` function as basically just being a helper function which creates the pool but that the caller is responsible for cleaning up the pool it receives when it is done using it.

`create-pool` could be called multiple times throughout the life of the service.  I'm not sure I'd want to have to store references for each of the pools which are created in the `jruby-pool-manager-service` so that they can be referenced for cleanup at shutdown.

The biggest issue I'd have with adding on the `stop` method is that its possible that a theoretical client may decide to hold one or more of its JRuby instances back until its own TK `stop` method is called, then the stop logic in the `jruby-pool-manager-service` could cause an indefinite hang.

I don't think any of the things I'm concerned about would affect any known use cases in Puppet Server so very possible that I'm overthinking this.  But I guess my preference for now would be to just do the extra cleanup in the common construct in the tests for now and maybe reconsider doing TK-380 later.  I can add a similar comment to TK-380 for posterity, too.

src/java/com/puppetlabs/jruby_utils/jruby/InternalScriptingContainer.java
camlow325: It's the declared return type for the [Class.forName()](https://docs.oracle.com/javase/7/docs/api/java/lang/Class.html#forName(java.lang.String,%20boolean,%20java.lang.ClassLoader)) method.  The <> part indicates that is a generic type.  The ? inside of it would signify that basically that the return type is expected to be a Class of a type extending from an `Object`.  I might have been able to do this as a `Class<JarIndex>` instead but `JarIndex` isn't public.

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
camlow325: Yeah, I can add something about JRuby 9k to the message and look into adding a test for it.  A non-mocked test would obviously only be able to assert the generation of the warning when the tests are running under JRuby 9k but that seems okay to me.

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
camlow325: Yeah, I struggled with that a little bit.  At this level, we try to coerce the compat-version from the config, if any, down to a value meeting the `SupportedJRubyCompatVersions`, which currently don't map to any values JRuby 9k would support.  I'll play around with this a bit and see if I can get the code that logs the full JRuby config to log the actual Ruby language version, even for 9k, where it outputs the `compat-version`.

src/java/com/puppetlabs/jruby_utils/jruby/InternalScriptingContainer.java
rlinehan: Ah cool, thanks for the explanation!

rlinehan:
👍 I think it would be good to leave a comment on TK-380 for posterity, but I think what you've said all makes sense and I'm good with it.

camlow325:
@rlinehan I added a commit, 867c9d3, onto the PR which I think addresses all of your previous feedback.  The correct Ruby language version should now be logged for the `compat-version` during JRuby config initialization, for both JRuby 1.7 and 9k.  I've also included a few additional tests to validate the warning messages and that JRuby 9k is actually in use when we expect it to be.

rlinehan:
👍 I think this all looks good to me. @jpinsonault you good if I merge?

PR #49, opened 2017-05-19
(SERVER-1815) Bump jruby-deps versions to 1.7.26-2 and 9.1.9.0-1

camlow325:
This commit bumps the jruby-deps version dependencies from 1.7.26-1 to
1.7.26-2 and from 9.1.8.0-1 to 9.1.9.0-1.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -21,7 +21,7 @@
                  [prismatic/schema]
                  [slingshot]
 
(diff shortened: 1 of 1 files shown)


camlow325:
This will fail Travis until after https://github.com/puppetlabs/jruby-deps/pull/8 is merged and jruby-deps 9.1.9.0-1 is released from it.  We'll need to wait for that before merging this one.

PR #52, opened 2017-05-25
(SERVER-1661) Bump JRuby to 1.7.27 and remove compat-version configurability

camlow325:
This PR bumps the jruby-deps dependency for 1.7 from 1.7.26-2 to
1.7.27-1.

This PR removes support for the compat-version setting from
jruby-utils.  The compat-version setting was introduced in jruby-utils
version 0.4.0 to allow for JRuby 1.7 to optionally run under Ruby
language version 2.0 instead of the default version, 1.9.  JRuby
1.7.27 effectively breaks the ability to use Ruby language version
2.0, however, due to a regression - see:
https://github.com/jruby/jruby/issues/4613 - and is unlikely to be
fixed in the future.  JRuby 9k also does not support compat-version
configurability.  Now that jruby-utils can be used with
JRuby 9k for Ruby language 2+ support, it is no longer valuable to be
able to set compat-version to 2 for use with JRuby 1.7. 

This PR also adds a helper method, jruby-version-info, for getting a
basic version info string from the underlying JRuby library.

### DIFF
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
@@ -16,7 +16,7 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Private
 
-(def default-jruby-compat-version
+(def default-jruby-1-7-compat-version
   "Default value for JRuby's 'CompatVersion' setting.  This value is only
   meaningful for JRuby 1.7.  For JRuby 9k, this will return `nil` because
   JRuby 9k effectively doesn't support configurable language compatibility
@@ -52,7 +52,7 @@
     (doto jruby
       (.setLoadPaths ruby-load-path)
       (.setCompileMode (get-compile-mode compile-mode)))
-    (when-let [compat-version default-jruby-compat-version]
+    (when-let [compat-version default-jruby-1-7-compat-version]
       (.setCompatVersion jruby compat-version))
     (initialize-scripting-container-fn jruby config)))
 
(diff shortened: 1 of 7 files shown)


camlow325:
This will fail Travis currently because the jruby-deps 1.7.26-1 release doesn't exist yet and so shouldn't be merged.  Putting it up for review only at this point.

camlow325:
jruby-deps 1.7.26-1 exists now and the PR is passing Travis so this should be good to review for merge now.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
```diff
@@ -16,6 +16,13 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Private
 
+(def default-jruby-compat-version
```
rlinehan: Would it be better to call this `default-jruby-1.7-compat-version` or something to indicate its only used for 1.7?

test/integration/puppetlabs/services/jruby_pool_manager/jruby_internal_test.clj
```diff
@@ -28,35 +28,18 @@
...
-                    :compat-version 2.0}))
+                   {:compile-mode :jit}))
           instance (jruby-internal/create-pool-instance! pool 0 config #())
           container (:scripting-container instance)]
       (try
         (is (= RubyInstanceConfig$CompileMode/JIT
             (.getCompileMode container)))
-        (is (.is2_0 (.getCompatVersion container)))
+        (if jruby-schemas/using-jruby-9k?
+          (is (.is1_9 (.getCompatVersion container))))
```
rlinehan: This seems... odd? Why are we testing that the compat version is 1.9 when we're using 9k? Or am I misunderstanding this test?

camlow325: oh, duh.  I wanted the reverse of that here, you're right.  I'll fix that up, thanks.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
camlow325: Yeah, I think that makes sense.  I'll change it to that.

camlow325: Changed this in 9ca965e.

test/integration/puppetlabs/services/jruby_pool_manager/jruby_internal_test.clj
camlow325: Fixed this in 9d3412a.

PR #56, opened 2017-12-07
(SERVER-1780) Enable jruby profiling

haus:


src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
```diff
@@ -162,6 +166,8 @@
   [config :- {schema/Keyword schema/Any}]
   (-> config
       (update-in [:compile-mode] #(keyword (or % default-jruby-compile-mode)))
+      (update-in [:profiling-mode] #(keyword (or % default-jruby-profiling-mode)))
+      (update-in [:profiler-output-file] #(or % nil))
```
haus: After talking this over with @lindboe I think I'm going to update the default here to a randomly named file that would be relative to the current working dir of the service, unless there are objections.

haus: With no file set the default behavior would be to dump profiling to stderr, which I think might go to different places on different platforms (probably the -daemon log on sysv platforms and journalctl on systemd platforms).

Magisus: +1 to the file in the working dir.

src/clj/puppetlabs/services/jruby_pool_manager/jruby_schemas.clj
```diff
@@ -62,7 +69,13 @@
         will be pooled.
 
     * :environment-vars - A map of environment variables and their values to be
-        passed through to the JRuby scripting container and visible to any Ruby code."
+        passed through to the JRuby scripting container and visible to any Ruby code.
+
+    * ::profiling-mode - The value to use for JRuby's ProfilerMode setting. Legal
```
Magisus: Is this supposed to have two colons in front of it, or just one like the others?

haus: just one. i presume i messed it up or cursive's refactoring did something wrong

src/clj/puppetlabs/services/jruby_pool_manager/jruby_schemas.clj
```diff
@@ -34,6 +34,13 @@
   "Schema defining the supported values for the JRuby CompileMode setting."
   (apply schema/enum supported-jruby-compile-modes))
 
+(def supported-jruby-profiling-modes
```
justinstoller: Not seeing this used elsewhere, is there a a reason to have it broken out of the schema?

haus: i copied what had been done for the `supported-jruby-compile-modes`

justinstoller: It looks like Chris was using that set in an integration test.

https://github.com/puppetlabs/jruby-utils/commit/9d44cfed213c387fabb041433cf6de7b8946060e#diff-78532c5c3713f4d2fe00eda939a3e93fR19

haus: (unit)

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
```diff
@@ -42,18 +46,38 @@
     :force RubyInstanceConfig$CompileMode/FORCE
     :off RubyInstanceConfig$CompileMode/OFF))
 
+(schema/defn ^:always-validate get-profiling-mode :- RubyInstanceConfig$ProfilingMode
+  [config-profiling-mode :- jruby-schemas/SupportedJRubyProfilingModes]
+  (case config-profiling-mode
```
Iristyle: 👍  Matches http://jruby.org/apidocs/org/jruby/RubyInstanceConfig.ProfilingMode.html

I know no Clojure - but... any way to automatically map the available constants in `RubyInstanceConfig` to keywords (I think that's what `:api` is?) rather than hard-coding the mapping?

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
```diff
@@ -2,16 +2,20 @@
   (:require [schema.core :as schema]
             [puppetlabs.services.jruby-pool-manager.jruby-schemas :as jruby-schemas]
             [clojure.tools.logging :as log]
-            [puppetlabs.i18n.core :as i18n])
+            [puppetlabs.i18n.core :as i18n]
+            [clojure.java.io :as io]
```
lindboe: nitpick: I try to keep requires in some sort of order (on previous teams we agreed on roughly alphabetical, what works best here tends to depend on whether editors manage this for you)

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
lindboe: We could automate that mapping assuming theres' a way to ask a Java class what its constants are (I don't know how to do that off the top of my head) and then stringify them and parse the strings. But I think that's a little more gnarly than just hardcoding them?

haus: Maybe. I might be able to grab the enum values and map them to keywords.

This seems to work:
```
(map keyword (map (comp clojure.string/lower-case str) (RubyInstanceConfig$ProfilingMode/values)))
(:off :api :flat :graph :html :json :service)
```

haus: `(map (comp keyword clojure.string/lower-case str) (RubyInstanceConfig$ProfilingMode/values))`

haus: I think then to grab the right constant from a given keyword we would do something like.

```(RubyInstanceConfig$ProfilingMode/valueOf (clojure.string/upper-case (name :off)))
=> #object[org.jruby.RubyInstanceConfig$ProfilingMode 0x4d53e068 "OFF"]
```

lindboe: Oh nice, that worked great

haus: (I was following the previous pattern used for compile-mode here)

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
haus: i can rearrange that.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
lindboe: The hardcoding is sorta useful because people can look into this code to see what values they can pick from

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
```diff
@@ -42,18 +46,38 @@
...
    config :- jruby-schemas/JRubyConfig]
-  (let [{:keys [ruby-load-path compile-mode lifecycle]} config
+  (let [{:keys [ruby-load-path compile-mode lifecycle profiling-mode profiler-output-file]} config
         initialize-scripting-container-fn (:initialize-scripting-container lifecycle)]
     (doto jruby
       (.setLoadPaths ruby-load-path)
       (.setCompileMode (get-compile-mode compile-mode)))
     (when-let [compat-version default-jruby-1-7-compat-version]
       (.setCompatVersion jruby compat-version))
+    (when (and profiler-output-file (not= :off profiling-mode))
```
lindboe: I would consider breaking this part out into its own function that takes a jruby and modifies it/logs as necessary? Just to keep this function simple

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
```diff
@@ -162,6 +167,8 @@
   [config :- {schema/Keyword schema/Any}]
   (-> config
       (update-in [:compile-mode] #(keyword (or % default-jruby-compile-mode)))
+      (update-in [:profiling-mode] #(keyword (or % default-jruby-profiling-mode)))
```
lindboe: Hm I know this is for consistency but I personally think it's fine to just use `:off` here and with compile mode

lindboe: Nitpick that this just made me realize though: if the JRuby setting is 'ProfilerMode', it would be more consistent for our setting to be called :profiler-mode

src/clj/puppetlabs/services/jruby_pool_manager/jruby_schemas.clj
```diff
@@ -62,7 +69,15 @@
...
     * :environment-vars - A map of environment variables and their values to be
-        passed through to the JRuby scripting container and visible to any Ruby code."
+        passed through to the JRuby scripting container and visible to any Ruby code.
+
+    * ::profiling-mode - The value to use for JRuby's ProfilerMode setting. Legal
+        values are `:api`, `:flat`, `:graph`, `:html`, `:json`, `:off`, and
+        `:service`. Defaults to `:off`.
+
+    * :profiler-output-file - A target file to direct profiler output to. If
+        not set defaults to a random file relative to the working directory
```
lindboe: comma after "set"

lindboe:
This LGTM, my comments are all purely suggestions

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
Magisus: All the docs for the Java functions had `ProfilingMode`, we actually went through and renamed it to that. Did you find the setting called `ProfilerMode` somewhere?

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
Magisus: I personally prefer hardcoding it for that reason, but I agree it's more brittle.

justinstoller: I think that's what the schema / `supported-jruby-profiling-modes` is valuable for.

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
lindboe: It just mentioned that in the docstring above for "default-jruby-profiling-mode", so I guess it was the docstring that was inconsistent?

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
haus: I'm going to leave the list for the schema, and update this section to remove the case

src/clj/puppetlabs/services/jruby_pool_manager/jruby_schemas.clj
haus: i can inline these if that still makes sense.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
haus: seems good

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
haus: I inlined the default and removed the def with the bad docstring

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
```diff
@@ -42,18 +49,39 @@
     :force RubyInstanceConfig$CompileMode/FORCE
     :off RubyInstanceConfig$CompileMode/OFF))
 
+(schema/defn ^:always-validate get-profiling-mode :- RubyInstanceConfig$ProfilingMode
+  [config-profiling-mode :- jruby-schemas/SupportedJRubyProfilingModes]
+  (RubyInstanceConfig$ProfilingMode/valueOf (cstring/upper-case (name config-profiling-mode))))
```
haus: cursive was very unhappy about this line but it works fine?

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
```diff
@@ -42,18 +49,39 @@
...
+
+(schema/defn ^:always-validate setup-profiling
```
haus: naming is hard, if this name seems off i'm happy to change it.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
pcarlisle: looks fine, what's wrong with it

haus: incorrect arity of 1 for `RubyInstanceConfig$ProfilingMode/valueOf`

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
```diff
@@ -1,17 +1,24 @@
 (ns puppetlabs.services.jruby-pool-manager.impl.jruby-internal
-  (:require [schema.core :as schema]
-            [puppetlabs.services.jruby-pool-manager.jruby-schemas :as jruby-schemas]
+  (:require [clj-time.core :as time-core]
+            [clj-time.format :as time-format]
+            [clojure.java.io :as io]
+            [clojure.string :as cstring]
```
justinstoller: I don't know if I've ever seen this import as anything other than `:as str`. What's the rationale for `cstring`?

test/integration/puppetlabs/services/jruby_pool_manager/jruby_internal_test.clj
```diff
@@ -31,16 +33,26 @@
...
+                   {:compile-mode :jit
+                    :profiler-output-file profiler-file
+                    :profiling-mode :flat}))
           instance (jruby-internal/create-pool-instance! pool 0 config #())
           container (:scripting-container instance)]
       (try
         (is (= RubyInstanceConfig$CompileMode/JIT
             (.getCompileMode container)))
+        (is (= RubyInstanceConfig$ProfilingMode/FLAT
+               (.getProfilingMode container)))
```
justinstoller: 💯

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
haus: `str` is already in `clojure.core` so I thought it better to avoid it

PR #58, opened 2017-12-13
(SERVER-1780) Insert hashCode into profiler file name

haus:
Previously when profiler-output-file was set, it would be the same for
each jruby in the pool, which is undesireable. This commit adds hashCode
to the filename to ensure that they are unique across jrubies.

### DIFF
--- a/test/integration/puppetlabs/services/jruby_pool_manager/jruby_internal_test.clj
+++ b/test/integration/puppetlabs/services/jruby_pool_manager/jruby_internal_test.clj
@@ -32,7 +32,7 @@
 
 (deftest settings-plumbed-into-jruby-container
   (testing "settings plumbed into jruby container"
-    (let [pool (JRubyPool. 1)
+    (let [pool (JRubyPool. 2)
           profiler-file (str (ks/temp-file-name "foo"))
           config (logutils/with-test-logging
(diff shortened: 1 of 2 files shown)


test/integration/puppetlabs/services/jruby_pool_manager/jruby_internal_test.clj
```diff
@@ -50,9 +52,12 @@
           (is (= CompatVersion/RUBY1_9 (.getCompatVersion container))
               "Unexpected default compat version configured for JRuby 1.7 container"))
         (finally
-          (.terminate container)))
+          (.terminate container)
+          (.terminate container-two)))
       ;; Because we add the current datetime to the filename we need to glob for it here.
```
lindboe: might want to update the comment to note the part about the hashcode too, since that will also be in the glob

lindboe:
LGTM, just a minor note

test/integration/puppetlabs/services/jruby_pool_manager/jruby_internal_test.clj
haus: good call.

haus: updated.

PR #59, opened 2018-01-05
(SERVER-2060) Splay JRuby instance flushing

justinstoller:
Previously all JRuby instances would be flushed after they reached
`max-requests-per-instance`. This could cause a thundering herd of
JRuby instance initialization which we've observed during
performance testing.

This patch creates a `splay-instance-flush` configuration option
(available via code for testing but not intended to be plumbed through
to the user). If this is true (the default value) when we first
initialize a JRuby Pool and fill it with instances each instance
receives an `initial-borrows` value that is a fraction of the
max-requests. If set, the `return-to-pool` method will prefer the
value of initial-borrows to that of max-requests to determine if
the instance should be flushed. Subsequent instance creations wil not
receive an initial-borrows value unless the entire pool is being
flushed and refilled.

### DIFF
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
@@ -144,30 +144,56 @@
                          (:id instance)))))
 
 (schema/defn ^:always-validate
+  initial-borrows-value :- (schema/maybe schema/Int)
+  [id :- schema/Int
+   total-instances :- schema/Int
+   total-borrows :- schema/Int
+   initial-jruby? :- schema/Bool]
+  (when initial-jruby?
+    (let [which-step (inc (mod id total-instances))
+          step-size (quot total-borrows total-instances)]
+      ;; If total-instances is larger than total-borrows then step-size will be 0
+      (if (= 0 step-size)
+        nil
+        (* step-size which-step)))))
+
+(schema/defn ^:always-validate
   create-pool-instance! :- JRubyInstance
   "Creates a new JRubyInstance and adds it to the pool."
-  [pool :- jruby-schemas/pool-queue-type
-   id :- schema/Int
-   config :- jruby-schemas/JRubyConfig
-   flush-instance-fn :- IFn]
-  (let [{:keys [ruby-load-path lifecycle]} config
-        initialize-pool-instance-fn (:initialize-pool-instance lifecycle)]
-    (when-not ruby-load-path
-      (throw (Exception.
-              (i18n/trs "JRuby service missing config value 'ruby-load-path'"))))
-    (log/infof (i18n/trs "Creating JRubyInstance with id {0}." id))
-    (let [scripting-container (create-scripting-container
-                               config)]
-      (let [instance (jruby-schemas/map->JRubyInstance
-                      {:scripting-container scripting-container
-                       :id id
-                       :internal {:pool pool
-                                  :max-borrows (:max-borrows-per-instance config)
-                                  :flush-instance-fn flush-instance-fn
-                                  :state (atom {:borrow-count 0})}})
-            modified-instance (initialize-pool-instance-fn instance)]
-        (.register pool modified-instance)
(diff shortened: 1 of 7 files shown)


src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
```diff
@@ -143,31 +143,57 @@
...
+  initial-borrows-value :- (schema/maybe schema/Int)
+  [id :- schema/Int
+   total-instances :- schema/Int
+   total-borrows :- schema/Int
+   initial-jruby? :- schema/Bool]
+  (when initial-jruby?
+    (let [which-step (inc (mod id total-instances))
+          step-size (quot total-borrows total-instances)]
+      ;; If total-instances is larger than total-borrows then step-size will be 0
+      (if (= 0 step-size)
```
haus: `if` => `when-not` and lose the nil branch?

justinstoller: I always forget the `not` variations.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
```diff
@@ -143,31 +143,57 @@
     (log/infof (i18n/trs "Cleaned up old JRubyInstance with id {0}."
                          (:id instance)))))
 
+(schema/defn ^:always-validate
+  initial-borrows-value :- (schema/maybe schema/Int)
```
haus: I love me some docstrings.

test/unit/puppetlabs/services/jruby_pool_manager/jruby_pool_test.clj
```diff
@@ -219,6 +220,54 @@
    (jruby-testutils/jruby-config {:max-active-instances max-instances
                                   :max-borrows-per-instance max-borrows})))
 
+(defn drain-and-refill
```
haus: Any reason not to put this with its friends in `jruby-testutils`?

haus: Also a docstring would help here.

test/unit/puppetlabs/services/jruby_pool_manager/jruby_pool_test.clj
```diff
@@ -219,6 +220,54 @@
...
+  [pool-context
+   instance-count]
+  (let [instances (jruby-testutils/drain-pool pool-context 2)
```
haus: 2 being hardcoded seems surprising. I'm also not sure what it is for, even having read the `drain-pool` function.

test/unit/puppetlabs/services/jruby_pool_manager/jruby_pool_test.clj
```diff
@@ -220,53 +220,50 @@
...
+           original-instances-surviving-first-drain (set/intersection first-ids second-ids)
+           new-instance-after-first-drain (set/difference second-ids first-ids)
+           original-instances-surviving-second-drain (set/intersection first-ids third-ids)
+           original-instances-surviving-third-drain (set/intersection first-ids fourth-ids)]
+       (testing "Initial instances are flushed each splay interval"
+         (is (= 2 (count original-instances-surviving-first-drain)))
+         (is (= 1 (count original-instances-surviving-second-drain)))
+         (is (= 0 (count original-instances-surviving-third-drain))))
+       (testing "New instances are flushed at max-borrows"
+         (is (seq (set/intersection new-instance-after-first-drain third-ids)))
```
haus: `not-empty` might be more readable here than `seq`, although the docs recommend using `(seq foo)` and not `(not (empty? foo))`. they seem to ignore `(not-empty foo)`.

justinstoller: Agreed on the readability part, I was applying the recommendation for `(not (empty? foo))`, but I'll go ahead and change that.

PR #61, opened 2018-01-24
(PUP-8226) Set JRuby KCode, source encoding, and external encoding

Magisus:
This commit adds a function that can configure the KCode, external
encoding, and source encoding of the JRuby scripting containers. We use
this function when initializing our JRuby instances to set these
encodings to UTF-8, to avoid parsing source files as US-ASCII, which can
cause issues with translation interpolation.

Note that this is only relevant in JRuby 1.7; in JRuby 9K, source files are parsed
as UTF-8 by default.

### DIFF
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
@@ -69,6 +69,26 @@
       (log/info
        (i18n/trs "Writing jruby profiling output to ''{0}''" real-profiler-output-file)))))
 
+(schema/defn ^:always-validate set-ruby-encoding :- jruby-schemas/ConfigurableJRuby
+  "Sets the K code, source encoding, and external encoding of the JRuby instance to
+  the supplied encoding."
+  [kcode :- KCode
+   jruby :- jruby-schemas/ConfigurableJRuby]
+  (let [encoding-string (str (.getEncoding kcode))]
+    ;; RubyInstanceConfig exposes the encoding setters directly, while the
+    ;; ScriptingContainer does not.
+    (if (instance? RubyInstanceConfig jruby)
+      (doto jruby
+        (.setKCode kcode)
+        (.setSourceEncoding encoding-string)
+        (.setExternalEncoding encoding-string))
+      (let [config (.getRubyInstanceConfig (.getProvider jruby))]
+        (doto config
+          (.setKCode kcode)
+          (.setSourceEncoding encoding-string)
+          (.setExternalEncoding encoding-string)))))
+  jruby)
+
 (schema/defn ^:always-validate init-jruby :- jruby-schemas/ConfigurableJRuby
   "Applies configuration to a JRuby... thing.  See comments in `ConfigurableJRuby`
   schema for more details."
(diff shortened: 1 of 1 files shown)


src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
```diff
@@ -69,6 +69,26 @@
...
 
+(schema/defn ^:always-validate set-ruby-encoding :- jruby-schemas/ConfigurableJRuby
+  "Sets the K code, source encoding, and external encoding of the JRuby instance to
+  the supplied encoding."
+  [kcode :- KCode
+   jruby :- jruby-schemas/ConfigurableJRuby]
+  (let [encoding-string (str (.getEncoding kcode))]
+    ;; RubyInstanceConfig exposes the encoding setters directly, while the
+    ;; ScriptingContainer does not.
+    (if (instance? RubyInstanceConfig jruby)
```
haus: Do you think it is worth being explicit here and using `cond` instead of `if` so that we can test for both `RubyInstanceConfig` and `ScriptingContainer`?

haus:
We talked in person about this, but I think it's worth breaking the function up into one that expects a `RubyInstanceConfig` and one that gets a `RubyInstanceConfig` as all of the calls are the same once a `RubyInstanceConfig` exists.

Magisus:
Updated.

Magisus:
Note, this proved really hard to test at the unit level, because the encoding bug only appears when parsing Ruby source files to execute, which is difficult to do in the context of the configured scripting container.

haus:
Nope. I totally didn't. My test passes, but doesn't fail without this change, so that's too bad.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
```diff
@@ -18,7 +18,8 @@
            (com.puppetlabs.jruby_utils.jruby InternalScriptingContainer
                                              ScriptingContainer)
            (org.jruby.runtime.profile.builtin ProfileOutput)
-           (java.io File)))
+           (java.io File)
+           (org.jruby.util KCode)))
```
justinstoller: nit: if we're changing this can we put the imports in alphabetical order?

Magisus: Updated

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
```diff
@@ -69,6 +70,26 @@
...
+  "Sets the K code, source encoding, and external encoding of the JRuby config to
+  the supplied encoding."
+  [kcode :- KCode
+   jruby :- RubyInstanceConfig]
+  (let [encoding-string (str (.getEncoding kcode))]
+    (doto jruby
+      (.setKCode kcode)
+      (.setSourceEncoding encoding-string)
+      (.setExternalEncoding encoding-string)))
+  jruby)
```
pcarlisle: `doto` returns its first argument so this is unnecessary

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
```diff
@@ -69,6 +70,26 @@
...
+
+(schema/defn ^:always-validate set-ruby-encoding :- jruby-schemas/ConfigurableJRuby
+  [kcode :- KCode
+   jruby :- jruby-schemas/ConfigurableJRuby]
+  (if (instance? RubyInstanceConfig jruby)
+    (set-config-encoding kcode jruby)
+    (set-config-encoding kcode (.getRubyInstanceConfig (.getProvider jruby))))
+  jruby)
```
pcarlisle: same here

PR #63, opened 2018-02-26
(SERVER-2130) Update default compile-mode to jit for jruby9k

haus:
In jruby 9k, to get comparable performance to 1.7, jit must be enabled
for compile-mode. This commit updates the default compile mode for jruby
to be jit when the version is 9k, but maintains the old default of off
for jruby 1.7.

### DIFF
--- a/test/integration/puppetlabs/services/jruby_pool_manager/jruby_internal_test.clj
+++ b/test/integration/puppetlabs/services/jruby_pool_manager/jruby_internal_test.clj
@@ -61,3 +61,19 @@
                                 profiler-files)]
         (is (= 2 (count profiler-files)))
         (is (not-empty (slurp real-profiler-file)))))))
+
+(deftest default-compile-mode
+  (testing "default compile-mode changes based on jruby version"
+    (let [pool (JRubyPool. 1)
+          config (logutils/with-test-logging
+                  (jruby-testutils/jruby-config {}))
+          instance (jruby-internal/create-pool-instance! pool 0 config #())
+          container (:scripting-container instance)]
+      (try
+        (if jruby-schemas/using-jruby-9k?
+          (is (= RubyInstanceConfig$CompileMode/JIT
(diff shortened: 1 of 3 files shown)


lindboe:
Going for this instead of a feature flag in the module, or in addition to?

haus:
I think the feature flag in the module will still be needed (to manage CodeCache and JRUBY_JAR settings), and we could explicitly set this as part of that work if desired.

test/integration/puppetlabs/services/jruby_pool_manager/jruby_internal_test.clj
```diff
@@ -62,3 +61,19 @@
...
+
+(deftest default-compile-mode
+  (testing "default compile-mode changes based on jruby version"
+    (let [pool (JRubyPool. 1)
+          config (logutils/with-test-logging
+                  (jruby-testutils/jruby-config {}))
+          instance (jruby-internal/create-pool-instance! pool 0 config #())
+          container (:scripting-container instance)]
+      (try
+        (if jruby-schemas/using-jruby-9k?
```
lindboe: I may not be understanding how this test works, but won't this only test one version of this conditional? I don't see how this test would get run twice.

haus: It doesn't get run twice, but there is a jruby9k profile that will also be run.

haus: (and travis runs both the standard 1.7 profile as well as the 9k profile https://github.com/puppetlabs/jruby-utils/blob/master/ext/travisci/test.sh#L8-L12)

lindboe: Ah I see, that travis info is helpful, thank you

PR #68, opened 2018-05-04
(SERVER-1777) Do not override JRuby ScriptingContainter.terminate

justinstoller:
Previously, we manually introspected JRuby's classloading and tempdir
handling to ensure that libraries copied to a tempdir when loaded would
be cleaned up on JRuby instance termination (upstream registers them as
deleteOnExit of the JVM).

With
https://github.com/jruby/jruby/commit/6493b3a93dcca2f9e8fac746cee8cf807594d161
JRuby now removes these temporary jars when the per-instance classloader
is terminated. This fix is available in JRuby 9.1.14.0 and later and
supersedes our code.


justinstoller:
I ran across this and figured I'd open a PR for it. I still need to run puppetserver against this and do some manual functional validation.

PR #69, opened 2018-11-02
(SERVER-2381) Use JRuby 9.2.0.0

justinstoller:


justinstoller:
I think the assumption that we'd branch from 1.x is a good one, I did not try this with 1.7 though I believe (from the ticket) it's unneeded.

justinstoller:
Travis is failing against a jdk7. I don't think that's a supported thing? Would anyone be opposed to me updating our travis matrix to test against 8 & 11 like we test for in the versions of server that will take this up?

Magisus:
That seems fine to me. Are there other consumers of this library that might care? (This seems like a more general relevant question: what else at Puppet uses JRuby besides puppetserver?)

justinstoller:
pebuildinfo says it goes into pe-console-services. Which surprises me (unless they're using it for testing), I'll look into their usage.

justinstoller:
Weird, lein won't even do a `lein deps` on jdk11 in travis. Locally I get:
```
09:11:37 jruby-utils justin (server2381 %) ‖ lein test :all
Tried to load org.apache.httpcomponents/httpcore version 4.1.2 but 4.4.4 was already loaded.
....
Tried to load commons-logging version 1.1.1 but 1.2 was already loaded.
Tried to load org.apache.maven/maven-model-builder version 3.0.4 but 3.5.0 was already loaded.
Tried to load org.apache.maven/maven-model version 3.0.4 but 3.5.0 was already loaded.
lein test puppetlabs.jruby-utils.slj4j-logger-test
lein test puppetlabs.jruby_utils.lockable-pool-test
lein test puppetlabs.services.jruby-pool-manager.jruby-agents-test
lein test puppetlabs.services.jruby-pool-manager.jruby-core-test
lein test puppetlabs.services.jruby-pool-manager.jruby-internal-test
Profiling enabled; ^C shutdown will now dump profile info
Profiling enabled; ^C shutdown will now dump profile info
lein test puppetlabs.services.jruby-pool-manager.jruby-interpreter-test
lein test puppetlabs.services.jruby-pool-manager.jruby-locking-test
lein test puppetlabs.services.jruby-pool-manager.jruby-pool-int-test
lein test puppetlabs.services.jruby-pool-manager.jruby-pool-test
lein test puppetlabs.services.jruby-pool-manager.jruby-service-test
lein test puppetlabs.services.jruby-pool-manager.jruby-testutils
Ran 62 tests containing 309 assertions.
0 failures, 0 errors.

09:12:15 jruby-utils justin (server2381 %) ‖ java -version
openjdk version "11.0.1" 2018-10-16
OpenJDK Runtime Environment 18.9 (build 11.0.1+13)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.1+13, mixed mode)
```

justinstoller:
wrong lein version, I think

justinstoller:
I seem to have angered the consistency gods. Depending which tab I'm in I see different commits/travis states for this PR. I'm going to stop messing with this and hope it eventually works itself out. All the changes should be up in my branch to get this green in travis.

I also checked and no one else internally is using this library.

project.clj
```diff
@@ -16,12 +16,13 @@
 
   :dependencies [[org.clojure/clojure]
                  [org.clojure/tools.logging]
+                 [org.clojure/tools.nrepl "0.2.13"]
```
Magisus: Why is this needed now?

justinstoller: It resolves a conflict between leiningen and clj-parent 2.x, I think every project that has gone to clj-parent 2.x has had to do this.

Magisus: Yeah I've seen that elsewhere, just think I missed what it was for. Thanks!

PR #71, opened 2019-08-19
(SERVER-2193) Add API for retrieving JRuby thread dumps

Sharpie:
This commit adds a `get-jruby-thread-dump` method that iterates through
the interpreters registered to a JRuby pool and returns a Ruby backtrace
from each one. Generation of the backtraces is done via the JRuby JMX
management interface and thus requires `-Djruby.management.enabled=true`
to be set in JAVA_ARGS.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -15,6 +15,7 @@
   :test-paths ["test/unit" "test/integration"]
 
   :dependencies [[org.clojure/clojure]
+                 [org.clojure/java.jmx]
                  [org.clojure/tools.logging]
 
                  [me.raynes/fs]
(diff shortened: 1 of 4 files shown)


Sharpie:
One test currently fails as JRuby backtraces from JMX are broken in 9.2.7.0. Fixed in 9.2.8.0:

https://github.com/jruby/jruby/commit/d2875a4

Magisus:
We're hoping to update to JRuby 9.2.8 shortly.

Magisus:
I have a PR for that update here https://github.com/puppetlabs/jruby-deps/pull/18, we're currently in the throes of a bunch of other dep bumps that might slow that down, but it should go in soon.

Sharpie:
With JRuby 9.2.8.0 in, tests now pass.

PR #75, opened 2019-08-27
Experiment with "reference pool"

Magisus:
This work creates a second implementation of a LockablePool, that hands out references to a single JRuby instance, instead of handing out multiple instances one at a time. This can be toggled via a `multithreaded` config value. It adds one additional setting, `max-concurrent-thread-count`, which controls how many references the ReferencePool may hand out at once.

It still needs more testing, partially in the form of a refactor of the existing tests, splitting some of them out because they apply only to the JRubyPool and not the ReferencePool paradigm (a good example of splayed flushing of instances) or vice versa. I also haven't done any high-level (e.g. in puppetserver) testing yet.

But before I do all that, I thought I'd put this up, so people can take a look at the approach and provide initial feedback on whether we think this is better than proceeding down the route started in https://github.com/puppetlabs/puppetserver/pull/2143. The semantics here are _almost_ right in most cases, and could probably be reworded in some places to reduce misleading descriptions and names. But it's also possible that this is conflating two ideas that are too dissimilar to warrant trying to jam them into one abstraction.

### DIFF
--- a/dev-resources/logback-test.xml
+++ b/dev-resources/logback-test.xml
@@ -7,7 +7,7 @@
 
     <!--<logger name="log4j.logger.org.eclipse.jetty.server" level="warn"/>-->
 
-    <root level="warn">
+    <root level="info">
         <appender-ref ref="STDOUT" />
     </root>
 </configuration>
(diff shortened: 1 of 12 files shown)


Magisus:
This now passes all of its tests. I'll have to completely redo the commits here to make them make sense. My next step is to try this in puppetserver.

Magisus:
Oh wait, apparently that work was on a copy of this branch. Going to clean up commits and open a new PR.

Magisus:
New PR: https://github.com/puppetlabs/jruby-utils/pull/77

PR #77, opened 2019-09-17
(SERVER-2529) Add ReferencePool

Magisus:
This adds a new implementation of the LockablePool interface,
that only maintains one JRuby instance and simply hands out multiple
references to it on a borrow request. The pool can be locked and
unlocked just like the JRubyPool.

The main semantic difference is that the pool can only be cleared
(and the instance unregistered) when all borrows have been returned.
This prevents us from deleting a JRuby instance out from under a
thread that is still using it.

Notable changes to the Clojure code surrounding the pool implemenation are:

1) Which pool is used can be toggled via the `jruby-puppet.multithreaded`
setting in the tk config.

2) The addition of a JRubyPool protocol that defines different methods
for clearing an individual instance and for clearing the whole pool.
This was necessary because the ReferencePool cannot meaningfully clear a
single instance -- clearing the instance is more or less equivalent to
clearing the whole pool, and therefore blocks on all the borrowed
references being returned. The new protocol handles this polymorphism.

### DIFF
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
@@ -5,19 +5,24 @@
             [puppetlabs.kitchensink.core :as ks]
             [puppetlabs.services.jruby-pool-manager.jruby-schemas :as jruby-schemas]
             [puppetlabs.i18n.core :as i18n]
-            [slingshot.slingshot :as sling])
+            [slingshot.slingshot :as sling]
+            [puppetlabs.services.protocols.jruby-pool :as pool-protocol])
   (:import (clojure.lang IFn IDeref)
-           (puppetlabs.services.jruby_pool_manager.jruby_schemas PoisonPill JRubyInstance)
+           (puppetlabs.services.jruby_pool_manager.jruby_schemas PoisonPill
+                                                                 JRubyInstance
+                                                                 ReferencePoolContext
+                                                                 JRubyPoolContext)
            (java.util.concurrent TimeUnit TimeoutException)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Private
 
+
 (schema/defn ^:always-validate
   next-instance-id :- schema/Int
   [id :- schema/Int
    pool-context :- jruby-schemas/PoolContext]
-  (let [pool-size (jruby-internal/get-pool-size pool-context)
+  (let [pool-size (jruby-internal/get-instance-count pool-context)
         next-id (+ id pool-size)]
     (if (> next-id Integer/MAX_VALUE)
       (mod next-id pool-size)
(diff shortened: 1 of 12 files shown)


src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -45,6 +50,92 @@
...
+  [pool-context instance cleanup-fn]
+  (let [pool-state (jruby-internal/get-pool-state-container pool-context)
+        pool (jruby-internal/get-pool pool-context)
+        config (:config pool-context)
+        _ (.releaseItem pool instance)]
+    (when (not (:flush-pending @pool-state))
+      ;; trigger a flush of the Jruby pool if one hasn't already
+      ;; been started
+      (try
+        (do
```
pcarlisle: you don't need a `do` within a `try`

Magisus:
For reference, this pre-commit squash: https://github.com/Magisus/jruby-utils/commits/ref-pool-messy

Magisus:
Some of the code could definitely be cleaned up, but I'm going to do some more functional testing first.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -45,6 +50,92 @@
...
+        (log/info (i18n/trs "Finished creating JRubyInstance 1 of 1"))
+        (swap! pool-state assoc :flush-pending false)
+        (catch Exception e
+          (.clear pool)
+          (.unlock pool)
+          (jruby-internal/insert-poison-pill pool e)
+          (throw (IllegalStateException.
+                   (i18n/trs "There was a problem flushing the pool.")
+                   e)))
+        (finally
```
pcarlisle: Why does this function always unlock the pool but `flush-reference-pool-instance!` leaves it locked if there's an error

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -45,6 +50,92 @@
...
+          (.releaseItem pool instance))
+        (.clear pool)
+        (when refill?
+          (jruby-internal/create-pool-instance! pool 1 (:config pool-context)
+                                                (partial send-flush-instance! pool-context)))
+        (log/info (i18n/trs "Finished creating JRubyInstance 1 of 1"))
+        (swap! pool-state assoc :flush-pending false)
+        (catch Exception e
+          (.clear pool)
+          (.unlock pool)
```
pcarlisle: this unlock is redundant with the one in the finally clause

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_pool_manager_core.clj
```diff
@@ -2,18 +2,24 @@
...
-    {:config config
-     :internal {:modify-instance-agent (jruby-agents/pool-agent shutdown-on-error-fn)
-                :pool-state (atom (jruby-internal/create-pool-from-config config))
-                :event-callbacks (atom [])}}))
+  (let [shutdown-on-error-fn (get-in config [:lifecycle :shutdown-on-error])
+        pool-state (if (:multithreaded config)
+                     (atom (jruby-internal/create-reference-pool-from-config config))
+                     (atom (jruby-internal/create-pool-from-config config)))
+        pool-context-internal {:modify-instance-agent (jruby-agents/pool-agent shutdown-on-error-fn)
+                               :pool-state            pool-state
```
pcarlisle: i know this is controversial historically, but i think we agreed eventually that it's better not to align the second column like this

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
pcarlisle: It's not clear to me how much the state matters when there's an error, but if it does it seems inconsistent

Magisus: I think they should both unlock. That's an oversight.

pcarlisle: You probably want to acquire the lock outside of the try, so you don't try to unlock in the case where you didn't successfully lock.

src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_agents.clj
```diff
@@ -45,6 +50,92 @@
...
+(schema/defn flush-reference-pool!
+  [pool-context
+   refill?
+   on-complete]
+  (let [pool-state (jruby-internal/get-pool-state-container pool-context)
+        pool (:pool (jruby-internal/get-pool-state pool-context))
+        cleanup-fn (get-in pool-context [:config :lifecycle :cleanup])]
+    (when (not (:flush-pending @pool-state))
+      (try
+        (swap! pool-state assoc :flush-pending true)
```
pcarlisle: This is racy since you test and set separately.

pcarlisle: Also it won't be set back in case of failure.

Magisus: Note to self: use `compare-and-set!` here

Magisus:
There are still enough issues with this (particularly in flushing semantics) that I've gone back to the drawing board (literally) and after talking with Justin, want to play around with adding a more robust pool abstraction above the Java locking queue implementation.

PR #78, opened 2019-10-07
(SERVER-2529) Clojure Reference Pool

Magisus:
This new attempt defines a Clojure protocol for interacting with the JRuby pool. This allows for some polymorphism of higher-level logic around the pool itself, which is particularly important for flushing with `max-requests-per-instance` (which involves the whole pool when using the ReferencePool versus just one instance in the original implementation) and for shutting down. It also allows for cleaner, easier-to-read logic in other places (like pool filling) that differ between the backing implementations.

### DIFF
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
@@ -46,7 +46,7 @@
     [pool-context instance]
     (when (jruby-schemas/jruby-instance? instance)
       (let [new-state (swap! (jruby-internal/get-instance-state-container instance)
-                             update-in [:borrow-count] inc)
+                             #(update-in % [:borrow-count] inc))
             {:keys [initial-borrows max-borrows pool]} (:internal instance)
             borrow-limit (or initial-borrows max-borrows)]
         (if (and (pos? borrow-limit)
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
@@ -52,5 +52,9 @@
                           (:id instance)
                           borrow-limit))
             (jruby-agents/send-flush-instance! this instance))
-          (.releaseItem pool instance))))))
+          (.releaseItem pool instance)))))
+
+  (flush-pool
+    [this]
+    (jruby-agents/flush-and-repopulate-pool! this)))
 
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
@@ -35,7 +35,9 @@
       (.unlock pool)))
 
   (borrow
-    [this])
+    [this]
+    (let [pool (jruby-internal/get-pool this)]
+      (.borrowItem pool)))
 
   (return
     [this instance]
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
@@ -50,7 +50,7 @@
                  (>= (:borrow-count new-state) borrow-limit))
           (do
             (log/info
-                (i18n/trs "Flushing JRubyInstance {0} because it has exceeded its borrow limit of ({1})"
+                (i18n/trs "Flushing JRubyInstance {0} because it has exceeded its borrow limit of {1}"
                           (:id instance)
                           borrow-limit))
             (jruby-agents/send-flush-instance! this instance))
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
@@ -1,0 +1,33 @@
+(ns puppetlabs.services.jruby-pool-manager.impl.instance-pool
+  (:require [puppetlabs.services.jruby-pool-manager.impl.jruby-agents :as jruby-agents]
+            [puppetlabs.services.protocols.jruby-pool :as pool-protocol])
+  (:import (puppetlabs.services.protocols.jruby_pool JRubyPool)
+           (puppetlabs.services.jruby_pool_manager.jruby_schemas InstancePool)))
+
+(extend-type InstancePool
+  pool-protocol/JRubyPool
+  (fill
+    [this]
+    (jruby-agents/send-prime-pool! this))
+
+  (shutdown
+    [this]
+    (jruby-agents/flush-pool-for-shutdown! this))
+
+  (lock
+    [this])
+
+  (unlock
+    [this])
+
+  (add-instance
+    [this instance])
+
+  (remove-instance
+    [this instance])
+
+  (borrow
+    [this])
+
+  (return
+    [this instance]))
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
@@ -1,7 +1,10 @@
 (ns puppetlabs.services.jruby-pool-manager.impl.instance-pool
   (:require [puppetlabs.services.jruby-pool-manager.impl.jruby-agents :as jruby-agents]
             [puppetlabs.services.protocols.jruby-pool :as pool-protocol]
-            [puppetlabs.services.jruby-pool-manager.impl.jruby-internal :as jruby-internal])
+            [puppetlabs.services.jruby-pool-manager.impl.jruby-internal :as jruby-internal]
+            [puppetlabs.services.jruby-pool-manager.jruby-schemas :as jruby-schemas]
+            [clojure.tools.logging :as log]
+            [puppetlabs.i18n.core :as i18n])
   (:import (puppetlabs.services.jruby_pool_manager.jruby_schemas InstancePool)))
 
 (extend-type InstancePool
@@ -35,6 +38,19 @@
     [this])
 
   (return
-    [_ instance]
-    (jruby-internal/return-to-pool instance)))
+    [this instance]
+    (when (jruby-schemas/jruby-instance? instance)
+      (let [new-state (swap! (jruby-internal/get-instance-state-container instance)
+                             update-in [:borrow-count] inc)
+            {:keys [initial-borrows max-borrows pool]} (:internal instance)
+            borrow-limit (or initial-borrows max-borrows)]
+        (if (and (pos? borrow-limit)
+                 (>= (:borrow-count new-state) borrow-limit))
+          (do
+            (log/info
+                (i18n/trs "Flushing JRubyInstance {0} because it has exceeded its borrow limit of ({1})"
+                          (:id instance)
+                          borrow-limit))
+            (jruby-agents/send-flush-instance! this instance))
+          (.releaseItem pool instance))))))
 
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
@@ -10,40 +10,40 @@
 (extend-type InstancePool
   pool-protocol/JRubyPool
   (fill
-    [this]
-    (let [modify-instance-agent (jruby-agents/get-modify-instance-agent this)]
+    [pool-context]
+    (let [modify-instance-agent (jruby-agents/get-modify-instance-agent pool-context)]
       (jruby-agents/send-agent modify-instance-agent
-                               #(jruby-agents/prime-pool! this))))
+                               #(jruby-agents/prime-pool! pool-context))))
 
   (shutdown
-    [this]
-    (jruby-agents/flush-pool-for-shutdown! this))
+    [pool-context]
+    (jruby-agents/flush-pool-for-shutdown! pool-context))
 
   (lock
-    [this]
-    (let [pool (jruby-internal/get-pool this)]
+    [pool-context]
+    (let [pool (jruby-internal/get-pool pool-context)]
       (.lock pool)))
 
   (lock-with-timeout
-    [this timeout time-unit]
-    (let [pool (jruby-internal/get-pool this)]
+    [pool-context timeout time-unit]
+    (let [pool (jruby-internal/get-pool pool-context)]
       (.lockWithTimeout pool timeout time-unit)))
 
   (unlock
-    [this]
-    (let [pool (jruby-internal/get-pool this)]
+    [pool-context]
+    (let [pool (jruby-internal/get-pool pool-context)]
       (.unlock pool)))
 
   (borrow
-    [this]
-    (jruby-internal/borrow-from-pool this))
+    [pool-context]
+    (jruby-internal/borrow-from-pool pool-context))
 
   (borrow-with-timeout
-    [this timeout]
-    (jruby-internal/borrow-from-pool-with-timeout this timeout))
+    [pool-context timeout]
+    (jruby-internal/borrow-from-pool-with-timeout pool-context timeout))
 
   (return
-    [this instance]
+    [pool-context instance]
     (when (jruby-schemas/jruby-instance? instance)
       (let [new-state (swap! (jruby-internal/get-instance-state-container instance)
                              update-in [:borrow-count] inc)
@@ -56,10 +56,10 @@
                 (i18n/trs "Flushing JRubyInstance {0} because it has exceeded its borrow limit of {1}"
                           (:id instance)
                           borrow-limit))
-            (jruby-agents/send-flush-instance! this instance))
+            (jruby-agents/send-flush-instance! pool-context instance))
           (.releaseItem pool instance)))))
 
   (flush-pool
-    [this]
-    (jruby-agents/flush-and-repopulate-pool! this)))
+    [pool-context]
+    (jruby-agents/flush-and-repopulate-pool! pool-context)))
 
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
@@ -1,6 +1,7 @@
 (ns puppetlabs.services.jruby-pool-manager.impl.instance-pool
   (:require [puppetlabs.services.jruby-pool-manager.impl.jruby-agents :as jruby-agents]
-            [puppetlabs.services.protocols.jruby-pool :as pool-protocol])
+            [puppetlabs.services.protocols.jruby-pool :as pool-protocol]
+            [puppetlabs.services.jruby-pool-manager.impl.jruby-internal :as jruby-internal])
   (:import (puppetlabs.services.jruby_pool_manager.jruby_schemas InstancePool)))
 
 (extend-type InstancePool
@@ -16,10 +17,19 @@
     (jruby-agents/flush-pool-for-shutdown! this))
 
   (lock
-    [this])
+    [this]
+    (let [pool (jruby-internal/get-pool this)]
+      (.lock pool)))
+
+  (lock-with-timeout
+    [this timeout time-unit]
+    (let [pool (jruby-internal/get-pool this)]
+      (.lockWithTimeout pool timeout time-unit)))
 
   (unlock
-    [this])
+    [this]
+    (let [pool (jruby-internal/get-pool this)]
+      (.unlock pool)))
 
   (add-instance
     [this instance])
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
@@ -32,7 +32,7 @@
       (.unlock pool)))
 
   (add-instance
-    [this instance])
+    [this id])
 
   (remove-instance
     [this instance])
@@ -41,4 +41,6 @@
     [this])
 
   (return
-    [this instance]))
+    [_ instance]
+    (jruby-internal/return-to-pool instance)))
+
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
@@ -31,12 +31,6 @@
     (let [pool (jruby-internal/get-pool this)]
       (.unlock pool)))
 
-  (add-instance
-    [this id])
-
-  (remove-instance
-    [this instance])
-
   (borrow
     [this])
 
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
@@ -36,8 +36,11 @@
 
   (borrow
     [this]
-    (let [pool (jruby-internal/get-pool this)]
-      (.borrowItem pool)))
+    (jruby-internal/borrow-from-pool this))
+
+  (borrow-with-timeout
+    [this timeout]
+    (jruby-internal/borrow-from-pool-with-timeout this timeout))
 
   (return
     [this instance]
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
@@ -1,14 +1,15 @@
 (ns puppetlabs.services.jruby-pool-manager.impl.instance-pool
   (:require [puppetlabs.services.jruby-pool-manager.impl.jruby-agents :as jruby-agents]
             [puppetlabs.services.protocols.jruby-pool :as pool-protocol])
-  (:import (puppetlabs.services.protocols.jruby_pool JRubyPool)
-           (puppetlabs.services.jruby_pool_manager.jruby_schemas InstancePool)))
+  (:import (puppetlabs.services.jruby_pool_manager.jruby_schemas InstancePool)))
 
 (extend-type InstancePool
   pool-protocol/JRubyPool
   (fill
     [this]
-    (jruby-agents/send-prime-pool! this))
+    (let [modify-instance-agent (jruby-agents/get-modify-instance-agent this)]
+      (jruby-agents/send-agent modify-instance-agent
+                               #(jruby-agents/prime-pool! this))))
 
   (shutdown
     [this]
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
@@ -12,7 +12,7 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; Private
 
-(schema/defn flush-pool
+(schema/defn flush-pool*
   "Flushes the pool, assuming it has already been locked by the calling function.
   Do not call this without first locking the pool, or the flush may never complete,
   since it requires that all references be returned to proceed."
@@ -47,7 +47,7 @@
       ;; Now that we've successfully acquired the lock, check the borrows again
       ;; to make sure the pool wasn't flushed while we were waiting.
       (when (max-borrows-exceeded @borrow-count max-borrows)
-        (flush-pool pool-context))
+        (flush-pool* pool-context))
       (finally
         (pool-protocol/unlock pool-context)))))
 
@@ -113,7 +113,7 @@
     [pool-context]
     (pool-protocol/lock pool-context)
     (try
-      (flush-pool pool-context)
+      (flush-pool* pool-context)
       (finally
         (pool-protocol/unlock pool-context)))))
 
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
@@ -2,7 +2,9 @@
   (:require [puppetlabs.services.protocols.jruby-pool :as pool-protocol]
             [puppetlabs.services.jruby-pool-manager.impl.jruby-agents :as jruby-agents]
             [puppetlabs.services.jruby-pool-manager.impl.jruby-internal :as jruby-internal]
-            [puppetlabs.services.jruby-pool-manager.jruby-schemas :as jruby-schemas])
+            [puppetlabs.services.jruby-pool-manager.jruby-schemas :as jruby-schemas]
+            [clojure.tools.logging :as log]
+            [puppetlabs.i18n.core :as i18n])
   (:import (puppetlabs.services.jruby_pool_manager.jruby_schemas ReferencePool)))
 
 (extend-type ReferencePool
@@ -71,11 +73,12 @@
             borrow-count (:borrow-count this)
             cleanup-fn (get-in this [:config :lifecycle :cleanup])
             old-instance (.borrowItem pool)
-            prev-id (:id old-instance)
+            id (inc (:id old-instance))
             _ (.releaseItem pool old-instance)]
         ;; This will block waiting for all borrows to be returned
         (jruby-internal/cleanup-pool-instance! old-instance cleanup-fn)
-        (jruby-agents/add-instance this (inc prev-id))
+        (jruby-agents/add-instance this id)
+        (log/info (i18n/trs "Finished creating JRuby instance with id {0}" id))
         (reset! borrow-count 0))
       (finally
         (pool-protocol/unlock this)))))
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
@@ -41,8 +41,11 @@
 
   (borrow
     [this]
-    (let [pool (jruby-internal/get-pool this)]
-      (.borrowItem pool)))
+    (jruby-internal/borrow-from-pool this))
+
+  (borrow-with-timeout
+    [this timeout]
+    (jruby-internal/borrow-from-pool-with-timeout this timeout))
 
   (return
     [this instance]
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
@@ -4,8 +4,26 @@
             [puppetlabs.services.jruby-pool-manager.impl.jruby-internal :as jruby-internal]
             [puppetlabs.services.jruby-pool-manager.jruby-schemas :as jruby-schemas]
             [clojure.tools.logging :as log]
-            [puppetlabs.i18n.core :as i18n])
-  (:import (puppetlabs.services.jruby_pool_manager.jruby_schemas ReferencePool)))
+            [puppetlabs.i18n.core :as i18n]
+            [schema.core :as schema])
+  (:import (puppetlabs.services.jruby_pool_manager.jruby_schemas ReferencePool
+                                                                 JRubyInstance)))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Private
+
+(schema/defn flush-if-at-max-borrows
+  [pool-context :- jruby-schemas/PoolContext
+   instance :- JRubyInstance]
+  (let [borrow-count (:borrow-count pool-context)
+        max-borrows (get-in instance [:internal :max-borrows])]
+    ;; If max-borrows is 0, never flush the instance
+    (when (and (pos? max-borrows)
+               (>= @borrow-count max-borrows))
+      (pool-protocol/flush-pool pool-context))))
+
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; ReferencePool definition
 
 (extend-type ReferencePool
   pool-protocol/JRubyPool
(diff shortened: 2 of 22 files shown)


Magisus:
This doesn't tackle any of the hard stuff yet, but I just wanted to get it up here. No need to review just yet.

src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
```diff
@@ -0,0 +1,62 @@
...
+    [this timeout time-unit]
+    (let [pool (jruby-internal/get-pool this)]
+      (.lockWithTimeout pool timeout time-unit)))
+
+  (unlock
+    [this]
+    (let [pool (jruby-internal/get-pool this)]
+      (.unlock pool)))
+
+  (borrow
```
Magisus: This currently isn't being called anywhere. There is (currently) no difference in borrow behavior between the two implementations. Also a lot of tests (in both this lib and in puppetserver) call `borrowItem` directly on the pool. But for consistency, this and its sister function `borrow-with-timeout` should probably be properly implemented and used.

src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
```diff
@@ -0,0 +1,86 @@
...
+        (swap! borrow-count inc)
+        ;; If max-borrows is 0, never flush the instance
+        (when (and (pos? max-borrows)
+                   (>= @borrow-count max-borrows)
+                   ;; If the pool is already locked, there's a good chance
+                   ;; a flush is already in progress, and we shouldn't queue
+                   ;; another one. If the pool is locked for some other reason,
+                   ;; we'll flush later. `borrow-count` does not get set back to
+                   ;; 0 until a flush succeeds.
+                   ;; THIS IS RACY
```
Magisus: But still helps prevent duplicate flushes some (most?) of the time. I'd appreciate thoughts on better ways to do this.

src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
```diff
@@ -0,0 +1,86 @@
...
+  (flush-pool
+    [this]
+    (pool-protocol/lock this)
+    (try
+      (let [pool (jruby-internal/get-pool this)
+            borrow-count (:borrow-count this)
+            cleanup-fn (get-in this [:config :lifecycle :cleanup])
+            old-instance (.borrowItem pool)
+            id (inc (:id old-instance))
+            _ (.releaseItem pool old-instance)]
```
Magisus: Don't really like this borrow-release block, but since we need a reference to an instance to get the idea, and no refs to instances can be held while flushing, it seemed necessary...

src/clj/puppetlabs/services/protocols/jruby_pool.clj
```diff
@@ -0,0 +1,26 @@
...
+  (lock
+    [this])
+
+  (lock-with-timeout
+    [this timeout time-unit])
+
+  (unlock
+    [this])
+
+  (borrow
```
Magisus: Should add a `borrow-with-timeout` function here to complete the interface.

Magisus:
Once integrated, this passes all of puppetserver's clojure tests with `multithreaded` turned off. I've filed some additional tickets to deal with the fallout when turning `multithreaded` on. Nothing too troubling.

src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
```diff
@@ -0,0 +1,86 @@
...
+            [clojure.tools.logging :as log]
+            [puppetlabs.i18n.core :as i18n])
+  (:import (puppetlabs.services.jruby_pool_manager.jruby_schemas ReferencePool)))
+
+(extend-type ReferencePool
+  pool-protocol/JRubyPool
+
+  (fill
+    [this]
+    (jruby-agents/add-instance this 1))
```
Magisus: This is synchronous, which is a departure from the old implementation, which spun off new threads to create instances. Since only one instance is being added, and no other pool operations can proceed while there is no instance, it seemed okay to do this, but I only did it to make things easier to test, which seems like a bad reason. Thoughts on this?

src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
```diff
@@ -0,0 +1,65 @@
...
+    [this]
+    (jruby-internal/borrow-from-pool this))
+
+  (borrow-with-timeout
+    [this timeout]
+    (jruby-internal/borrow-from-pool-with-timeout this timeout))
+
+  (return
+    [this instance]
+    (when (jruby-schemas/jruby-instance? instance)
```
mwaggett: When would this not be a jruby instance?

src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
```diff
@@ -0,0 +1,65 @@
...
+      (let [new-state (swap! (jruby-internal/get-instance-state-container instance)
+                             update-in [:borrow-count] inc)
```
mwaggett: Should `update-in [:borrow-count] inc` be in its own set of parentheses? Maybe it doesn't need to be, but I think that would help me read this method? Unless I'm misunderstanding what `swap!` does..

Magisus: The lack of parens is [required](https://clojuredocs.org/clojure.core/swap!). `swap!` takes several args, first the atom being updated, then a function, then each of the args to that function, as many as there are. In this case, the function is `update-in`, which itself takes the map (passed in by `swap!`), the list of keys pointing to the value to be udpated, then the function to apply to the value of that key.

It's honestly pretty confusing. But in general, no parens are ever optional in Clojure. If they're there, they tend to be there for a reason, same if they're missing. There's very little syntax leeway in the language.

src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
Magisus: It might be a pill, if this thread tried to borrow when a pill had been inserted. I actually added this when I found a test that verified that when a pill is inserted, no error is thrown. The Java `return` method will error if the thing being returned is not the registered instance, see https://github.com/puppetlabs/jruby-utils/pull/78/files#diff-4486a0d17a50859d82ce51b87b75c9d5R256. In practice, the pill being returned is most likely also the one that the pool is holding onto, but in this test, it was a new one that hadn't actually been inserted into the pool yet.

src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
mwaggett: ahh good to know, thanks!

src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
mwaggett: so to make sure I'm understanding - the only reason you don't also call `return` on pills is so you can avoid that error in a test where the pill was never registered?

src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
```diff
@@ -0,0 +1,65 @@
...
+            {:keys [initial-borrows max-borrows pool]} (:internal instance)
+            borrow-limit (or initial-borrows max-borrows)]
+        (if (and (pos? borrow-limit)
+                 (>= (:borrow-count new-state) borrow-limit))
+          (do
+            (log/info
+                (i18n/trs "Flushing JRubyInstance {0} because it has exceeded its borrow limit of ({1})"
```
mwaggett: Why parentheses around the number? I would expect either
`Flushing JRubyInstance {0} because it has exceeded its borrow limit of {1}` or
`Flushing JRubyInstance {0} because it has exceeded its borrow limit ({1})`

src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
Magisus: This mirrors the original implementation, and it looks like the [simple commit](https://github.com/puppetlabs/jruby-utils/commit/4bf607037e00e74b3be2185e10b04e842e3c5ba6) that added that is part of a very fraught PR: https://github.com/puppetlabs/jruby-utils/pull/34. Such a case might never arise in the wild (that something that is not an instance or the registered pill might be returned), but I wasn't sure, so I just added the same logic. If we think it's safe to remove here (and let the logic in the Java handle it), then we should also remove it (and the weird test) from the original implementation.

src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
Magisus: Agreed, I'm going to drop the parens

src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
mwaggett: I'm generally a big fan of removing things, but don't necessarily have strong feeling right here right now. Mostly wanted to make sure I was understanding things.

src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
```diff
@@ -0,0 +1,89 @@
...
+                   ;; another one. If the pool is locked for some other reason,
+                   ;; we'll flush later. `borrow-count` does not get set back to
+                   ;; 0 until a flush succeeds.
+                   ;; THIS IS RACY
+                   (not (.isLocked pool)))
+          (jruby-agents/send-agent modify-instance-agent #(pool-protocol/flush-pool this))))))
+
+  (flush-pool
+    [this]
+    (pool-protocol/lock this)
```
Magisus: This library supports a `flush-timeout` config option that we need to support here. Found via a test in puppetserver that verifies it.

src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
justinstoller: I think this is idiomatic (and the way it was originally written [here](https://github.com/puppetlabs/jruby-utils/pull/78/files#diff-63fb42862df919c7a769bbd0da1c3ee3L349)), though could also be:
```clj
(swap! ... #(update-in % [:borrow-count] inc))
```

Magisus: Oh huh. I do like that better, I'll update.

src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
justinstoller: I didn't totally follow this thread, but what I got from it is that this protocol method `return` can take a jruby-instance or poison-pill, but none of this logic is applicable to poison-pills, so if poison-pill (eg not a jruby-instance) be a no-op. Is that accurate?

src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
pcarlisle: If you acquire the lock around this part then you won't retrieve borrow count until the flush is done and when you get it it will be near 0 and should therefore not trigger a new flush. (It's usually an error to check a lock instead of just acquiring it.) I think it would be ideal actually if this whole check occurred inside the function that the agent calls.

src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
pcarlisle: probably ok?

justinstoller: Is it possible to insert some kind of flow control as an initialize-instance-fn in the test vs making this synchronous?

fwiw, I'm unconcerned about the pool operations being synchronous so much as the calling puppetserver start up code becoming synchronous.

src/clj/puppetlabs/services/protocols/jruby_pool.clj
```diff
@@ -0,0 +1,29 @@
+(ns puppetlabs.services.protocols.jruby-pool)
+
+(defprotocol JRubyPool
```
justinstoller: These should have comments describing the inputs and expected behavior from a users perspective. (eg clarify https://github.com/puppetlabs/jruby-utils/pull/78/files#r353362968)

src/clj/puppetlabs/services/jruby_pool_manager/jruby_schemas.clj
```diff
@@ -127,8 +139,7 @@
                'JRubyInstanceState))
 
 (def JRubyPuppetInstanceInternal
-  {:flush-instance-fn IFn
-   :pool pool-queue-type
+  {:pool pool-queue-type
```
justinstoller: Do we still need this? Can it instead be a reference to the "owning" pool-context? Would that make some of the signature changes unnecessary?

Magisus: I can look into how that is used. I know that we get a reference to the pool itself (the Java class) in a lot of places, but I think we pretty much always pull it out of `PoolState` above.

I think doing that would require more signature changes, while also still requiring the existing ones.. but maybe I'm missing which ones you're referring to?

justinstoller: nvm, I was conflating a couple of things.

src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
Magisus: If I'm understanding this right, putting the whole check inside the agent-called function is better because the agent will ensure that any other action the agent is performing (e.g. an in-progress flush) has completed before it calls the function again, ensuring that when we check the conditions, they will represent the post-flush state. Is that accurate? The docs on agents are a little hard to parse...

Magisus:
I updated both `flush` and `fill` for the reference pool to run via the agent, which means the strategy of using watchers in the flush tests no longer works. I'm going to look into other ways to write those tests next. Suggestions welcome.

src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
Magisus: Update: we should check the borrows both before and after locking, and if they're back to 0 after we've acquired the lock, we should skip flushing and just unlock immediately.

src/java/com/puppetlabs/jruby_utils/pool/ReferencePool.java
```diff
@@ -0,0 +1,526 @@
...
+    // borrowed from the pool.
+    private final Condition lockAvailable = borrowLock.newCondition();
+
+    // Condition signaled when an element has been added into the queue.
+    // Awaited when a request has been made to borrow an item but no elements
+    // currently exist in the queue.
+    private final Condition borrowsAvailable = borrowLock.newCondition();
+
+    // Condition signaled when the pool has been unlocked.  Awaited when a
+    // request has been made to borrow an reference or lock the pool but the pool
```
mwaggett: an -> a

src/java/com/puppetlabs/jruby_utils/pool/ReferencePool.java
```diff
@@ -0,0 +1,526 @@
...
+        this.maxBorrowCount = maxBorrows;
+        this.currentBorrowCount = new AtomicInteger(0);
+    }
+
+    @Override
+    public void register(E e) {
+        final ReentrantLock lock = this.borrowLock;
+        lock.lock();
+        try {
+            if (instance != null) {
```
mwaggett: I would find it easier to read if this was `this.instance` instead of just `instance` - it makes it clearer to me where `instance` is coming from. thought I definitely won't block on this - I don't know what conventions have been established in this repo already.

src/java/com/puppetlabs/jruby_utils/pool/ReferencePool.java
```diff
@@ -0,0 +1,526 @@
...
+        } finally {
+            lock.unlock();
+        }
+    }
+
+    /**
+     * Unregisters the JRuby instance. Blocks waiting for all borrows of the
+     * instance to be returned before clearing it out. In this pool implementation,
+     * `clear` and `unregister` are aliases of one another, since clearing the pool
+     * is the same as clearing the on active instance.
```
mwaggett: on -> one

src/java/com/puppetlabs/jruby_utils/pool/ReferencePool.java
```diff
@@ -0,0 +1,526 @@
...
+
+    @Override
+    public E borrowItemWithTimeout(long timeout, TimeUnit unit) throws
+            InterruptedException {
+        E item = null;
+        final ReentrantLock lock = this.borrowLock;
+        long remainingMaxTimeToWait = unit.toNanos(timeout);
+
+        // `borrowLock.lockInterruptibly()` is called here as opposed to just
+        // `borrowLock.borrowLock` to follow the pattern that the JDK's
```
mwaggett: Should these be `lock.<blah>`?
And should the second one be `lock.lock()`? 
I'm looking at the methods above.

src/java/com/puppetlabs/jruby_utils/pool/ReferencePool.java
```diff
@@ -0,0 +1,526 @@
...
+     * inserted
+     */
+    @Override
+    public void releaseItem(E e) {
+        final ReentrantLock lock = this.borrowLock;
+        lock.lock();
+        try {
+            if (e != this.pill) {
+                if (!isRegistered(e)) {
+                    String errorMsg = "The item being released is not registered with the pool";
```
mwaggett: Would it be useful to include what the item is in the error message?

src/java/com/puppetlabs/jruby_utils/pool/ReferencePool.java
```diff
@@ -0,0 +1,526 @@
...
+      Set<E> registered = new CopyOnWriteArraySet<E>();
+      if (instance != null) {
+          registered.add(instance);
+      }
+      return registered;
+    }
+
+    private void freePoolLock() {
+        poolLockThread = null;
+        // Need to use 'signalAll' here because there might be multiple
```
mwaggett: Why do we ever want to _not_ use `signalAll`?

mwaggett: nvm, I see more info in a comment below

src/java/com/puppetlabs/jruby_utils/pool/ReferencePool.java
```diff
@@ -0,0 +1,526 @@
...
+    private boolean isPoolLockHeld() {
+        return poolLockThread != null;
+    }
+
+    private boolean isPoolLockHeldByCurrentThread(Thread currentThread) {
+        return poolLockThread == currentThread;
+    }
+
+    private boolean isPoolLockHeldByAnotherThread(Thread currentThread) {
+        return (poolLockThread != null) && (poolLockThread != currentThread);
```
mwaggett: curious if it would make sense to use `isPoolLockHeld` and `isPoolLockHeldByCurrentThread` in this method. though maybe it's really just tomato tomato.

test/unit/puppetlabs/jruby_utils/lockable_ref_pool_test.clj
```diff
@@ -0,0 +1,638 @@
...
+  (deref ref 10000 :timed-out))
+
+(defn create-empty-pool
+  ([] (create-empty-pool 10))
+  ([maxBorrows]
+   (ReferencePool. maxBorrows)))
+
+(defn create-populated-pool
+  ([] create-populated-pool 10)
+  ([size]
```
mwaggett: I think this should be called `maxBorrows` as well

test/unit/puppetlabs/jruby_utils/lockable_ref_pool_test.clj
```diff
@@ -0,0 +1,638 @@
...
+    (is (not (.isLocked pool)))
+    (let [lock-thread (future (deliver future-started? true)
+                              (.lock pool)
+                              (deliver lock-acquired? true)
+                              @unlock?
+                              (.unlock pool)
+                              true)]
+      @future-started?
+      (testing "pool.lock() blocks until all instances are returned to the pool"
+        (is (not (realized? lock-acquired?)))
```
mwaggett: Seems like this could race with the `lock-thread`..

Magisus: Ah, I know why it doesn't. In the let at the top, we borrow 3 of the instances. The lock can't proceed while the instances are borrowed.

Magisus:
I think I've fixed the race condition in the tests (haven't seen it again after a dozen or so runs, used to happen every two or three times) in a way that actually makes sense.

I also filed https://tickets.puppetlabs.com/browse/SERVER-2688 for what I think is a transient unrelated to this work.

src/java/com/puppetlabs/jruby_utils/pool/ReferencePool.java
Magisus: So there are a _lot_ of places in this file that use fields without prefixing `this`. In my IDE, the syntax highlighting makes fields a different color than the local variables, but I notice that my color scheme in vim does not (and neither does GitHub). I don't have a strong preference here, but many of the other places will get significantly more verbose if we consistently add `this` everywhere.

I'd be interested to hear if others have a policy for approaching this beyond "any time there's potential shadowing" (most common in constructors in my experience). It's unclear to me what heuristic the original Java authors in this repo were following.

src/java/com/puppetlabs/jruby_utils/pool/ReferencePool.java
Magisus: I'm gonna change them to just reference the method name.

src/java/com/puppetlabs/jruby_utils/pool/ReferencePool.java
Magisus: What sort of information do you think would be useful here? It's kind of ugly to print an object ID, maybe just adding the type information? Not sure if this could get ugly if it's a type coming in from Clojure.

src/java/com/puppetlabs/jruby_utils/pool/ReferencePool.java
Magisus: Yeah we should use those methods. I like names.

test/unit/puppetlabs/jruby_utils/lockable_ref_pool_test.clj
mwaggett: so then what happens when `deliver lock-acquired? true` is called and the instances aren't released yet?

Magisus: `(.lock pool)` will block until all of the instances are released, then proceed to acquire the lock. So we won't deliver the promise until after the lock has succeeded.

src/java/com/puppetlabs/jruby_utils/pool/ReferencePool.java
mwaggett: I'm not sure, since I don't really know what object info is available or what a user might find actionable. Mostly just thinking generally/hypothetically about error messages and imagining myself seeing that message and then wondering "what item??" but I don't really know when/if that might come up.

test/unit/puppetlabs/jruby_utils/lockable_ref_pool_test.clj
mwaggett: ahhh right right, gotcha

mwaggett: oh wow that is literally what it says it's testing 🤦‍♀

Magisus: Still good to understand where all the actual blocking and proceeding is taking place! It's definitely not obvious if you're not super familiar with the pool's methods.

test/unit/puppetlabs/jruby_utils/lockable_ref_pool_test.clj
```diff
@@ -0,0 +1,638 @@
...
+      (is (true? (timed-deref lock-thread))
+          "timed out waiting for the lock thread to finish")
+      (is (true? (timed-deref blocked-borrow-thread-borrowed?))
+          "timed out waiting for the borrow to complete")
+      (is (true? (timed-deref blocked-borrow-thread))
+          "timed out waiting for the borrow thread to complete")
+      (is (not (.isLocked pool))))))
+
+(deftest pool-lock-reentrant-for-borrow-from-locking-thread
+  (testing "the thread that holds the pool lock may borrow instances while holding the lock"
```
mwaggett: Why do we want to allow this?

test/unit/puppetlabs/jruby_utils/lockable_ref_pool_test.clj
```diff
@@ -0,0 +1,638 @@
...
+                              @unlock?
+                              (.unlock pool)
+                              true)]
+      @blocked-borrow-thread-started?
+      @lock-thread-started?
+      (is (not (realized? blocked-borrow-thread-borrowed?)))
+      (let [start (System/currentTimeMillis)]
+        (while (and (not (.isLocked pool))
+                    (< (- (System/currentTimeMillis) start) 10000))
+          (Thread/yield)))
```
mwaggett: could be useful to comment why we're waiting here

test/unit/puppetlabs/jruby_utils/lockable_ref_pool_test.clj
Magisus: I actually make use of it several times in the ReferencePool implementation of the Clojure protocol. It's okay because the thread that locks the pool does so in order to have complete control over all of its instances, so it can do things like modify and delete them. Since the only way to get a reference to an instance is by borrowing it, we need to allow borrows to happen by the locking thread while the pool is locked.

A great example of this is when we are deleting an instance. One part of deleting an instance is to clean up the Ruby scripting container associated with it. For the ReferencePool in particular, where the instance can be borrowed by lots of threads, we need lock the pool in order to make sure no one else is borrowing the instance, then get a reference to it (by borrowing it ourselves) and clean up its scripting container, then return our reference back to the pool. See https://github.com/puppetlabs/jruby-utils/pull/78/files#diff-1392e2eb7a0c370666b5d0ec72185b55R66-R74 for an example. Note that we actually return before we clean up here. That's both safe -- because we've prevented anyone else from getting ahold of it because we've locked the pool --  and required, because another thing `cleanup` does is unregister the instance, which we only allow if the pool's current borrow count is 0. I don't particularly _like_ this logic, but it's kind of the only option given that the cleanup code is in Clojure. (Maybe we could re-evaluate that? I think Justin was musing about the possibility at some point)

test/unit/puppetlabs/jruby_utils/lockable_ref_pool_test.clj
```diff
@@ -0,0 +1,638 @@
...
+                                  true)]
+        @lock-thread-started-1?
+        @lock-thread-started-2?
+        (is (not (realized? lock-thread-locked-1?)))
+        (is (not (realized? lock-thread-locked-2?)))
+        (.unlock pool)
+        (is (true? (timed-deref lock-thread-1))
+            "timed out waiting for first lock thread to finish")
+        (is (true? (timed-deref lock-thread-2))
+            "timed out waiting for second lock thread to finish"))
```
mwaggett: Will these two threads race to acquire the lock once the main thread unlocks? Maybe that doesn't even matter? I guess I'm wondering why two lock threads are even involved in this test?

test/unit/puppetlabs/jruby_utils/lockable_ref_pool_test.clj
```diff
@@ -0,0 +1,638 @@
...
+        lock-thread-locked? (promise)
+        lock-thread (future (deliver lock-thread-obj (Thread/currentThread))
+                            (.lock pool)
+                            (deliver lock-thread-locked? true))]
+    (testing (str "write locker's thread can be interrupted while waiting for "
+                  "instances to be returned")
+      (.interrupt @lock-thread-obj)
+      (is (thrown? ExecutionException (timed-deref lock-thread)))
+      (is (not (realized? lock-thread-locked?))))
+      (is (not (.isLocked pool)))
```
mwaggett: nitpick, so feel free to ignore, but I would like if these two were switched to match the order of the `.lock` and `deliver`.

test/unit/puppetlabs/jruby_utils/lockable_ref_pool_test.clj
Magisus: I'm not sure why there are multiple, though the test description makes it seem like that was very intentional. They might race in that it's undetermined who will acquire the lock first, but since both threads also immediately unlock, whoever gets it first will give it up right away, allowing the other one to proceed. So by the time the second test gets evaluated, both are guaranteed to be done (either the first will complete immediately, or it will block till Thread 2 gives up the lock and Thread 1 acquires it, and then succeed that way).

test/unit/puppetlabs/jruby_utils/lockable_ref_pool_test.clj
```diff
@@ -0,0 +1,638 @@
...
+      ; Only to show that it does not error
+      (is (nil? (.releaseItem pool first-pill)))
+      (is (thrown-with-msg?
+           IllegalArgumentException
+           #"The item being released is not registered with the pool"
+           (.releaseItem pool second-pill)))))
+
+  (testing "releasing a jruby not registered with the pool errors"
+    (let [pool (create-populated-pool 1)
+          not-in-pool-instance "I was never registered"]
```
mwaggett: not sure if we care, but earlier in this test, we use `str()` instead of just a straight-up quoted string. we might want consistency?

test/unit/puppetlabs/jruby_utils/lockable_ref_pool_test.clj
```diff
@@ -0,0 +1,638 @@
...
+
+  (testing "a blocked .lock call throws an InterruptedException once a pill is inserted"
+    (let [pool (create-populated-pool 1)
+          pill "pillful ignorance"]
+      ; Make it so the pool is not full
+      (.borrowItem pool)
+
+      ; Exceptions thrown from the future will be returned as InterruptedException,
+      ; so we can't use thrown-with-msg?. We'll catch it, return it instead of
+      ; throwing it, and inspect it manually below
```
mwaggett: This could be clarified, imo. When I read this, it looks like the `InterruptedException` is the reason we can't use `thrown-with-msg?`, but the reason is actually because it's getting thrown from a future.

test/unit/puppetlabs/jruby_utils/lockable_ref_pool_test.clj
```diff
@@ -0,0 +1,638 @@
...
+      (let [blocked-lock-future (future (try (.lock pool)
+                                             (catch InterruptedException e
+                                               e)))]
+        ; The future's thread will take the lock, and then block waiting for
+        ; either the pool to fill up, or a pill to be inserted
```
mwaggett: Is this saying that the waiting is blocked, or is the "waiting for..." part describing what it's blocking on?

Magisus: The latter. It's blocking because it's waiting for one of those things to happen.

test/unit/puppetlabs/jruby_utils/lockable_ref_pool_test.clj
mwaggett: yeah, I think that's the conclusion I came to, but was still trying to understand the value of testing with two threads.

test/unit/puppetlabs/jruby_utils/lockable_ref_pool_test.clj
mwaggett: mkay, that's what I figured. I think the comment could be clarified  - perhaps "block **on** waiting . . . "?

mwaggett: or maybe "block (waiting for . . . )"?

src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
```diff
@@ -31,7 +39,44 @@
...
 
   (return
-    [this instance]))
+    [this instance]
+    (let [pool (jruby-internal/get-pool this)
+          borrow-count (:borrow-count this)
+          max-borrows (get-in instance [:internal :max-borrows])
+          modify-instance-agent (jruby-agents/get-modify-instance-agent this)]
+      (.releaseItem pool instance)
+      (swap! borrow-count inc)
```
mwaggett: perhaps there's a less-overloaded / clearer term than `borrow-count` here. maybe like `total-borrows-before-flush` or something like that..

mwaggett: or maybe `total-requests-per-instance`, similar to `max-requests-per-instance` like the commit says

test/unit/puppetlabs/services/jruby_pool_manager/jruby_ref_pool_test.clj
```diff
@@ -1,2 +1,107 @@
...
+                                  :max-borrows-per-instance max-borrows})))
+
+(deftest more-borrows-than-mrpi
+  (testing "flushing due to max borrows succeeds when we've over-borrowed")
+  (jruby-testutils/with-pool-context
+    pool-context
+    jruby-testutils/default-services
+    ;; We can check out the instance more times than `max-borrows` and return
+    ;; them each in sequence, and nothing blocks or fails
+    (jruby-testutils/jruby-config {:max-active-instances 4
```
mwaggett: This should call `jruby-test-config`.

test/unit/puppetlabs/services/jruby_pool_manager/jruby_ref_pool_test.clj
```diff
@@ -1,2 +1,107 @@
...
+        (jruby-core/return-to-pool pool-context instance :test [])
+        (let [instance (jruby-core/borrow-from-pool pool-context :test [])]
+          (is (= id (:id instance)))
+          (jruby-core/return-to-pool pool-context instance :test [])))))
+  (testing "JRubyInstance is flushed after exceeding max borrows"
+    (jruby-testutils/with-pool-context
+      pool-context
+      jruby-testutils/default-services
+      (jruby-test-config 2)
+      (jruby-testutils/wait-for-jrubies-from-pool-context pool-context)
```
mwaggett: I'm a little confused as to why this is used here and nowhere else.

Magisus: This test wound up getting deleted, but I'm thinking maybe it shouldn't have.

test/unit/puppetlabs/services/jruby_pool_manager/jruby_ref_pool_test.clj
```diff
@@ -12,41 +12,48 @@
                                   :multithreaded true
                                   :max-borrows-per-instance max-borrows})))
 
+(deftest borrow-while-no-instances-available-test
+  (testing "when all instances are in use, borrow blocks until an instance becomes available"
+    (let [pool-size 2]
+      (jruby-testutils/with-pool-context
+        pool-context
+        jruby-testutils/default-services
+        (jruby-testutils/jruby-config {:max-active-instances pool-size
```
mwaggett: Is this intentionally missing `:max-borrows-per-instance`?

mwaggett: if not, seems like we should just use `jruby-test-config`

test/unit/puppetlabs/services/jruby_pool_manager/jruby_ref_pool_test.clj
```diff
@@ -12,41 +12,48 @@
...
+              ;; return an instance to the pool
+              (jruby-core/return-to-pool pool-context
+                                          (first drained-instances)
+                                          :borrow-with-no-free-instances-test
+                                          [])
+
+              ;; now the borrow can complete
+              (is (some? @borrow-instance)))
+            (finally
+              (jruby-testutils/fill-drained-pool pool-context drained-instances))))))))
```
mwaggett: Why do we need to refill after the test?

Magisus: Because we need to make sure that we can shutdown as the last step of the macro, so that all the JRuby containers get shutdown properly (see https://github.com/puppetlabs/jruby-utils/commit/b08a9b0fb8c6df62a11da98e97dec9b10ef289ec). The shutdown can't proceed if we're holding onto any references.

test/unit/puppetlabs/services/jruby_pool_manager/jruby_ref_pool_test.clj
Magisus: Just kidding, we're still functionally testing the same scenario without it.

src/clj/puppetlabs/services/jruby_pool_manager/impl/instance_pool.clj
```diff
@@ -46,7 +46,7 @@
     [pool-context instance]
     (when (jruby-schemas/jruby-instance? instance)
       (let [new-state (swap! (jruby-internal/get-instance-state-container instance)
-                             update-in [:borrow-count] inc)
+                             #(update-in % [:borrow-count] inc))
```
mwaggett: 💯 this is much clearer to me than how it was previously written

src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
```diff
@@ -72,27 +101,19 @@
...
-            cleanup-fn (get-in pool-context [:config :lifecycle :cleanup])
-            old-instance (.borrowItem pool)
-            id (inc (:id old-instance))
-            _ (.releaseItem pool old-instance)]
-        ;; This will block waiting for all borrows to be returned
-        (jruby-internal/cleanup-pool-instance! old-instance cleanup-fn)
-        (jruby-agents/add-instance pool-context id)
-        (log/info (i18n/trs "Finished creating JRuby instance with id {0}" id))
-        (reset! borrow-count 0))
+      (flush-pool pool-context)
```
mwaggett: How does this know that `flush-pool` doesn't refer to itself, i.e. this function isn't recursive, it's calling the other `flush-pool` above?

test/unit/puppetlabs/services/jruby_pool_manager/jruby_ref_pool_test.clj
```diff
@@ -60,12 +60,15 @@
     (jruby-testutils/with-pool-context
       pool-context
       jruby-testutils/default-services
-      (jruby-test-config 2)
+      (jruby-testutils/jruby-config {:max-active-instances 1
```
mwaggett: This could still call `jruby-test-config`

test/unit/puppetlabs/jruby_utils/lockable_ref_pool_test.clj
```diff
@@ -0,0 +1,630 @@
...
+      ; inserting the pill should unblock the promise
+      (.insertPill pool pill)
+      ; borrow finishes and gives us back the pill
+      (let [future-result (timed-deref blocked-borrow)]
+        (is (identical? future-result pill)))))
+
+  (testing "second insert doesn't change the pill"
+    (let [pool (create-populated-pool 1)
+          first-pill (str "pill clinton")
+          second-pill (str "pillary clinton")]
```
mwaggett: might wanna un-`str()`-ify these too

Magisus:
Discovered an edge-casey race condition around shutdown, going to fix it before merge this.

test/unit/puppetlabs/services/jruby_pool_manager/jruby_ref_pool_test.clj
```diff
@@ -131,3 +133,36 @@
             pill (ShutdownPoisonPill. pool)]
         ; Returning a pill should be a noop
         (jruby-core/return-to-pool pool-context pill :test [])))))
+
+(deftest shutdown-prevents-flush
+  (testing "Triggering a flush after shutdown has been requested does not break shutdown"
+    (let [config (jruby-test-config 0 2)
+          ;; Set up the pool manually
```
mwaggett: Could be good to say *why* you're doing it manually.

Magisus: Good point. It was a weird thing actually.

tvpartytonight:
typing at pool

tvpartytonight:
I have nothing more to say on this PR except that it was very easy to follow the commit messages to see the progression. Good jorb!

PR #82, opened 2019-12-12
(SERVER-2685) Respect `flush-timeout` setting in ReferencePool

mwaggett:


Magisus:
This looks good so far!

src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
```diff
@@ -41,15 +43,21 @@
...
+      (pool-protocol/lock-with-timeout pool-context flush-timeout TimeUnit/MILLISECONDS)
+      (try
+        ;; Now that we've successfully acquired the lock, check the borrows again
+        ;; to make sure the pool wasn't flushed while we were waiting.
+        (when (max-borrows-exceeded @borrow-count max-borrows)
+          (flush-pool* pool-context))
+        (finally
+          (pool-protocol/unlock pool-context)))
+      (catch TimeoutException e
+        (log/info (i18n/trs "An attempt to lock the JRubyPool failed with a timeout"))))))
```
Magisus: Can we put some context here about why we were trying to lock? Maybe saying something about how we were trying to flush because we'd hit the max borrows per instance, and we'll try again?

mwaggett: For sure. I was trying to keep the message the same as the one for the other pool, but I definitely agree that more info would be useful.

test/unit/puppetlabs/services/jruby_pool_manager/jruby_ref_pool_test.clj
```diff
@@ -118,6 +125,40 @@
...
+      pool-context
+      jruby-testutils/default-services
+      ;; Set max-instances higher than max-borrows in order to over-borrow
+      (jruby-test-config 1 2 {:flush-timeout 1})
+      ;; Borrow instance1 so we can trigger a refresh when we return it
+      ;; Borrow instance2 to prevent lock from being acquired when we return instance1
+      (let [instance1 (pool-protocol/borrow pool-context)
+            instance2 (pool-protocol/borrow pool-context)
+            flush-complete (add-watch-for-flush-complete pool-context)]
+        ;; Return to trigger flush, which should lock since instance2 is still borrowed
```
Magisus: 'lock' -> 'block'

test/unit/puppetlabs/services/jruby_pool_manager/jruby_ref_pool_test.clj
```diff
@@ -118,6 +125,40 @@
...
+        (pool-protocol/return pool-context instance1)
+        (is (not (realized? flush-complete)))
+        ;; Return instance2 so that shutdown can proceed (otherwise this hangs)
```
Magisus: I think this comment might be misleading now, since we're blocking on checking for the flush to complete below. It should probably reference the flush instead of shutdown.

mwaggett: hmm I'm starting to wonder if this is really testing what we want to test.. 
I feel like this isn't proving that the flush timed out.. in theory it could just be waiting, right?

Magisus: That's a good point. I wonder if we can check for logging...?

test/unit/puppetlabs/services/jruby_pool_manager/jruby_ref_pool_test.clj
```diff
@@ -118,6 +126,43 @@
...
+      (let [instance1 (pool-protocol/borrow pool-context)
+            instance2 (pool-protocol/borrow pool-context)
+            flush-complete (add-watch-for-flush-complete pool-context)]
+        (logutils/with-test-logging
+          ;; Return to trigger flush, which should block since instance2 is still borrowed
+          (pool-protocol/return pool-context instance1)
+          (is (not (realized? flush-complete)))
+          ;; Wait for 2 seconds to make sure the flush really times out before proceeding
+          (Thread/sleep 2000)
+          (is (logged? "Max borrows reached, but JRubyPool could not be flushed because lock could not be acquired. Will try again later." :warn))
```
mwaggett: okay, I updated this test to check the logs for the timeout warning.

PR #83, opened 2020-01-07
(SERVER-2681) Add `availableJrubies` method

mwaggett:
This commit adds a method for getting the number of free Jrubies in the pool.
This is the same as `currentSize` for the existing instance pool, but for the
reference pool (multithreaded) this will either be 1 (if nothing is using the
pool) or 0. This ensures that the `num-free-jrubies` metric (which uses
`free-instance-count`) is accurate for both pool types.

### DIFF
--- a/src/java/com/puppetlabs/jruby_utils/pool/LockablePool.java
+++ b/src/java/com/puppetlabs/jruby_utils/pool/LockablePool.java
@@ -122,7 +122,12 @@ public interface LockablePool<E> {
      */
     int currentSize();
 
-   /**
+    /**
+     * Returns the number of available (i.e. unused) JRuby instances in the pool.
+     */
+    int availableJRubies();
+
+    /**
     * Lock the pool. This method should make the following guarantees:
     *
     *  a) blocks until all registered elements are returned to the pool
--- a/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
@@ -72,7 +72,7 @@
   "Returns the number of JRubyInstances available in the pool."
   [pool :- jruby-schemas/pool-queue-type]
   {:post [(>= % 0)]}
-  (.currentSize pool))
+  (.availableJRubies pool))
 
 (schema/defn ^:always-validate
   get-instance-state :- jruby-schemas/JRubyInstanceState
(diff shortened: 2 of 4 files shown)


justinstoller:
I didn't follow the `free-instance-count` part, but the code changes here seem sane. 👍

mwaggett:
ahh, I can try to clarify

mwaggett:
@justinstoller better?

Magisus:
We realize that this metric probably isn't super meaningful for the ReferencePool, and probably will suggest that people move to look at the "free workers" metric we're thinking of adding, which will represent unused request processing capacity for both pools (instances for the old impl, threading capacity for the new one). But this seemed to be the most accurate way to interpret the `free-instances` function for the new pool.

Unless you think something else makes more sense?

src/java/com/puppetlabs/jruby_utils/pool/LockablePool.java
```diff
@@ -122,7 +122,12 @@ E borrowItemWithTimeout(long timeout, TimeUnit unit)
      */
     int currentSize();
 
-   /**
+    /**
+     * Returns the number of jruby instances in the pool.
```
Magisus: "...number of unused jruby instances..."

mwaggett: whoops, will fix

src/java/com/puppetlabs/jruby_utils/pool/LockablePool.java
```diff
@@ -122,7 +122,12 @@ E borrowItemWithTimeout(long timeout, TimeUnit unit)
      */
     int currentSize();
 
-   /**
+    /**
+     * Returns the number of available (i.e. unused) Jruby instances in the pool.
```
mwaggett: @Magisus better?

src/java/com/puppetlabs/jruby_utils/pool/LockablePool.java
```diff
@@ -122,7 +122,12 @@ E borrowItemWithTimeout(long timeout, TimeUnit unit)
...
+     */
+    int availableJrubies();
```
Magisus: I think we usually capitalize the 'r' in 'JRuby' as well. Let's be consistent.

mwaggett: @Magisus I fixed the comment - did you want me to fix the functions too? I wasn't sure.

Magisus: Yes please, that was the one I was most interested in, since it's harder to change later.

mwaggett: okay done!

Magisus:
So I guess we'll just be able to use `currentSize` to back the `free-workers` metric when we add that?

mwaggett:
yeah, that sounds right

PR #85, opened 2020-01-13
(SERVER-2715) Record worker-id when borrowing pool resources

mwaggett:
This commit adds a `:worker-id` event field to be sent to the metrics service.
For the instance pool, this is the instance id we've already been using. For
the new reference pool, this is the thread id. This will allow us to reason
about pool usage in similar ways for both pool types.

### DIFF
--- a/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
@@ -200,8 +200,8 @@
    reason :- schema/Any
    event-callbacks :- [IFn]]
   (let [requested-event (jruby-events/instance-requested event-callbacks reason)
-        instance (pool-protocol/borrow pool-context)]
-    (jruby-events/instance-borrowed event-callbacks requested-event instance)
+        [instance worker-id] (pool-protocol/borrow pool-context)]
+    (jruby-events/instance-borrowed event-callbacks requested-event instance worker-id)
     instance))
 
 ;; TODO: consider adding a second arity that allows for passing in a
@@ -219,10 +219,10 @@
    event-callbacks :- [IFn]]
   (let [timeout (get-in pool-context [:config :borrow-timeout])
         requested-event (jruby-events/instance-requested event-callbacks reason)
-        instance (pool-protocol/borrow-with-timeout
-                   pool-context
-                   timeout)]
-    (jruby-events/instance-borrowed event-callbacks requested-event instance)
+        [instance worker-id] (pool-protocol/borrow-with-timeout
+                          pool-context
+                          timeout)]
+    (jruby-events/instance-borrowed event-callbacks requested-event instance worker-id)
     instance))
 
 (schema/defn ^:always-validate
@@ -232,8 +232,8 @@
    instance :- jruby-schemas/JRubyInstanceOrPill
    reason :- schema/Any
    event-callbacks :- [IFn]]
-  (jruby-events/instance-returned event-callbacks instance reason)
-  (pool-protocol/return pool-context instance))
+  (let [worker-id (pool-protocol/return pool-context instance)]
+    (jruby-events/instance-returned event-callbacks instance reason worker-id)))
 
 (schema/defn ^:always-validate
   flush-pool!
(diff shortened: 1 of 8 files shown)


mwaggett:
Tested by running `lein test :only puppetlabs.services.jruby.jruby-metrics-service-test/borrow-timeout-test` with https://github.com/puppetlabs/puppetserver/pull/2264 - previously there were two failures, now there are none!

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
```diff
@@ -200,8 +200,8 @@
    reason :- schema/Any
    event-callbacks :- [IFn]]
   (let [requested-event (jruby-events/instance-requested event-callbacks reason)
-        instance (pool-protocol/borrow pool-context)]
-    (jruby-events/instance-borrowed event-callbacks requested-event instance)
+        [instance worker-id] (pool-protocol/borrow pool-context)]
+    (jruby-events/instance-borrowed event-callbacks requested-event instance worker-id)
     instance))
```
Magisus: Oh, perfect, this keeps the API change contained 👍

src/clj/puppetlabs/services/protocols/jruby_pool.clj
```diff
@@ -25,14 +25,15 @@
...
 
   (borrow-with-timeout
     [pool-context timeout]
-    "Returns a reference to a JRuby instance. Will block if the pool is locked or no instances
-    are available, timing out when the supplied number of milliseconds has elapsed.")
+    "Returns a reference to a JRuby instance and a worker id (instance id or thread id).
+    Will block if the pool is locked or no instances are available, timing out when the
+    supplied number of milliseconds has elapsed.")
 
   (return
```
Magisus: Can you add something to the doc string here that this will now return the id of the thing being returned?

Magisus:
It might be worth adding some tests around the thread-id stuff, specific to the reference pool.

mwaggett:
I'll take a look at what testing makes sense.

src/clj/puppetlabs/services/protocols/jruby_pool.clj
```diff
@@ -25,18 +25,20 @@
...
     [pool-context instance]
-    "Releases a held reference to a JRuby instance back to the pool. If `max-requests-per-instance`
+    "Releases a held reference to a JRuby instance back to the pool and returns the worker id
+    (instance id or thread id) for the thing being returned. If `max-requests-per-instance`
```
mwaggett: @Magisus How does this sound?

test/unit/puppetlabs/services/jruby_pool_manager/jruby_ref_pool_test.clj
```diff
@@ -23,6 +23,18 @@
...
+  (testing "worker id is the same at borrow time and return time"
+    (jruby-testutils/with-pool-context
+      pool-context
+      jruby-testutils/default-services
+      (jruby-test-config 2)
+      (let [[instance borrowed-id] (pool-protocol/borrow pool-context)
+            returned-id (pool-protocol/return pool-context instance)]
+        (is (= (.getId (Thread/currentThread)) borrowed-id))
+        (is (= (.getId (Thread/currentThread)) returned-id))
+        (is (= borrowed-id returned-id))))))
```
Magisus: Probably don't need this last check now?

mwaggett: eh yeah, I suppose

src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
```diff
@@ -219,10 +219,10 @@
    event-callbacks :- [IFn]]
   (let [timeout (get-in pool-context [:config :borrow-timeout])
         requested-event (jruby-events/instance-requested event-callbacks reason)
-        instance (pool-protocol/borrow-with-timeout
-                   pool-context
-                   timeout)]
-    (jruby-events/instance-borrowed event-callbacks requested-event instance)
+        [instance worker-id] (pool-protocol/borrow-with-timeout
+                          pool-context
```
justinstoller: weird spacing

mwaggett: what should this align with?

mwaggett: I guess I can make it like how it was previously

Magisus:
This will need some follow-on work in puppetserver to actually fix the logging, but I'm going to merge this to make development on that part easier.

PR #86, opened 2020-01-17
(SERVER-2679) Ensure return callback runs before actual return

mwaggett:
This commit puts the `instance-returned` callback before the actual `return`
again in the public-facting `return-to-pool` method. Previously, if a request
was waiting for a new instance, calling `return` first could trigger a new
borrow so quickly that the borrow event callback would trigger before the return
event callback. For the instance pool, because instances ids are reused, instead
of 1. removing the instance from the `borrowed-instances` metric and then 2.
adding it back with the new borrow event, we ended up 1. updating the
`borrowed-instances` with the new event and then 2. removing the instance from
`borrowed-instances`, which caused the metric to lose track of a whole instance.

### DIFF
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
@@ -106,17 +106,19 @@
     (let [pool (jruby-internal/get-pool pool-context)]
       (.unlock pool)))
 
+  (worker-id
+    [pool-context instance]
+    (.getId (Thread/currentThread)))
+
   (borrow
     [pool-context]
-    (let [instance (jruby-internal/borrow-from-pool pool-context)
-          worker-id (.getId (Thread/currentThread))]
-      [instance worker-id]))
+    (let [instance (jruby-internal/borrow-from-pool pool-context)]
+      [instance (pool-protocol/worker-id pool-context instance)]))
 
   (borrow-with-timeout
     [pool-context timeout]
-    (let [instance (jruby-internal/borrow-from-pool-with-timeout pool-context timeout)
-          worker-id (.getId (Thread/currentThread))]
-      [instance worker-id]))
+    (let [instance (jruby-internal/borrow-from-pool-with-timeout pool-context timeout)]
+      [instance (pool-protocol/worker-id pool-context instance)]))
 
   (return
     [pool-context instance]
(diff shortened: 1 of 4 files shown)


mwaggett:
confirmed with https://github.com/puppetlabs/puppetserver/pull/2264 that the failures I was seeing are now gone 🎉

src/clj/puppetlabs/services/jruby_pool_manager/impl/reference_pool.clj
```diff
@@ -131,7 +133,7 @@
           (jruby-agents/send-agent modify-instance-agent
                                    #(flush-if-at-max-borrows pool-context instance)))
         ;; Return the worker-id, to be used in metrics and event logging
-        (.getId (Thread/currentThread)))))
+        (pool-protocol/worker-id pool-context instance))))
```
Magisus: I think we might not need to do this anymore, if you're calling `worker-id` explicitly in the return event registering?

mwaggett: This is helpful in testing that the worker-id is the same at borrow-time and return-time.

Though the real reason I kept it was to change as little as possible.. 😁

Magisus: Okay I think it's fine for now then.

src/clj/puppetlabs/services/protocols/jruby_pool.clj
```diff
@@ -23,6 +23,10 @@
     [pool-context]
     "Unlocks the JRuby pool, allowing borrows to proceed.")
 
+  (worker-id
+    [pool-context instance]
+    "Returns a reference to a worker id (instance id or thread id).")
```
Magisus: Can you rephrase this a little to say "Returns the worker id for a given instance"? It's not really a reference we're returning, and since that word is used so much with the new pool, I'd rather not overload it.

src/clj/puppetlabs/services/protocols/jruby_pool.clj
```diff
@@ -23,6 +23,10 @@
     [pool-context]
     "Unlocks the JRuby pool, allowing borrows to proceed.")
 
+  (worker-id
+    [pool-context instance]
+    "Returns the worker id for given instance (instance id or thread id).")
```
mwaggett: @Magisus updated

PR #91, opened 2020-05-27
(SERVER-2793) Disable invokedynamic for yields

pcarlisle:
This is on by default as of JRuby 9.2.11.0. It appears to be related to
stack overflows that we've started seeing, so turn it off for now.

### DIFF
--- a/src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj
@@ -101,6 +101,7 @@
       (.setCompileMode (get-compile-mode compile-mode)))
     (set-ruby-encoding KCode/UTF8 jruby)
     (setup-profiling jruby profiler-output-file profiling-mode)
+    (System/setProperty "jruby.invokedynamic.yield" "false")
     (initialize-scripting-container-fn jruby config)))
 
 (schema/defn ^:always-validate empty-scripting-container :- ScriptingContainer


justinstoller:
Makes sense but I would have expected this around https://github.com/puppetlabs/jruby-utils/blob/master/src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj#L92

I don't know that using the linked code is better, but it might be best to keep the same kind of code all together.

pcarlisle:
It looks like the jruby has already been instantiated in that block though. Unfortunately this setting just gets read at some point when the jruby is created and there isn't a method to change it on one.

justinstoller:
We depend on the lazy instantiation of the Ruby runtime (here: [ScriptingContainer.java#L458-L459](https://github.com/jruby/jruby/blob/master/core/src/main/java/org/jruby/embed/ScriptingContainer.java#L458-L459))  to configure the ScriptingContainer after it's created but before we run any ruby code.
For example here is the KCode function setting a ScriptingContainer's InstanceConfig after it's been initialized: [jruby_internal.clj#L89-L90](https://github.com/puppetlabs/jruby-utils/blob/master/src/clj/puppetlabs/services/jruby_pool_manager/impl/jruby_internal.clj#L89-L90)

pcarlisle:
ok i'll test that

pcarlisle:
ok it did seem to work

justinstoller:
Sorry, I wrote the above with the assumption that every thing defined as a setting was settable via RubyInstanceConfig. But looking through the JRuby code base I don't actually see an accessor for it??

justinstoller:
Given we have to set a system property directly I'm happy with either doing it next to the other configuration on every JRuby instantiation or once by itself at startup in the service, sorry.

PR #93, opened 2020-12-09
(maint) Use JRuby 9.2.14.0

justinstoller:


justinstoller:
btw, after this we should move `master` to `main`

mwaggett (approved):

PR #94, opened 2021-03-01
(SERVER-1243) Error when configured pool size is less than 1

Magisus:
It is not meaningful to create a JRuby pool with less than one instance,
because then any attempt to borrow an instance will hang. This commit
causes an error to be thrown if the user attempts to configure
`max-active-instances` to a value less than 1.


Magisus:
Putting this up so I don't forget about it. Could use a test.

PR #96, opened 2021-09-27
(SERVER-3098) Log pool lock events at INFO level

Sharpie:
The JRuby pool lock is used to gain exclusive access to the JRuby
pool. When requested, threads will be prevented from borrowing
a JRuby instance until all instances are returned to the pool and
the lock is released.

This behavior gives the JRuby lock lifecyle of request, acquire, release
an extremely critical impact on JRuby throughput. This change increses
the log statements around these lifecycle events to INFO level so
that post-mortem debugging is possible when services using the JRuby
pool are not logging at DEBUG level.

### DIFF
--- a/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
+++ b/src/clj/puppetlabs/services/jruby_pool_manager/jruby_core.clj
@@ -255,11 +255,11 @@
   [pool-context :- jruby-schemas/PoolContext
    reason :- schema/Any
    event-callbacks :- [IFn]]
-  (log/debug (i18n/trs "Acquiring lock on JRubyPool..."))
+  (log/info (i18n/trs "Acquiring lock on JRubyPool..."))
   (jruby-events/lock-requested event-callbacks reason)
   (pool-protocol/lock pool-context)
   (jruby-events/lock-acquired event-callbacks reason)
-  (log/debug (i18n/trs "Lock acquired")))
+  (log/info (i18n/trs "Lock acquired")))
 
 (schema/defn ^:always-validate
   lock-pool-with-timeout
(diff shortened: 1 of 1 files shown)


mwaggett:
This seems reasonable to me, as long as we're not concerned about this cluttering up the logs. 

Do we consider logging to be an API that we shouldn't break? If so, would any change be considered a breakage or only a change that would make the log messages _less_ visible (the opposite of what this PR does, so this would be fine)?

justinstoller:
It feels like adding 3 log lines per request is a lot. I spun up a repl and got 1 log line for a catalog compile per agent run. There's probably more logging in PE by default, but this would quadruple the number of logs per catalog in that case.

Would it be better to configure our product to run with the `puppetlabs.services.jruby-pool-manager.jruby-core` at debug level by default (it looks like these are the only messages in this namespace).

Magisus:
This is for the whole pool lock, right? So I think it would only log when file-sync locks it for a code-deploy, not on every catalog request.

justinstoller:
Oh yeah, lock, not borrow. I'm more okay with this then. Though it still seems like something better handled in a logback config.

Magisus:
I'm fine with either approach. Is your thought that in general we try to hide the pool's workings (by logging at debug), and this would sorta break that expectation?

This seems like a pretty important message: halting the whole pool's throughput is kind of a big deal. I'm actually a little surprised we weren't logging this at info before tbh.

Sharpie:
This could be tweaked via logback, but lots of existing installations out there have customized logback configurations that won't be updated by a package upgrade. This logging is critical enough that we should make sure the change is taken up by existing installations.

This shouldn't be noticeable to open source users as PE File Sync is the primary consumer of this locking API. On the PE side, there are maybe "dozens" of deployments per day, so a couple hundred more log lines should not make much of an impact on the signal to noise ratio.

Magisus (approved):

mwaggett (approved):

PR #112, opened 2023-10-10
(maint) Update jruby 9.3.11

justinstoller:
Depends on release jruby-deps from https://github.com/puppetlabs/jruby-deps/pull/51 both require https://github.com/puppetlabs/ci-job-configs/pull/9183


justinstoller:
Discussed in Slack that these updates are largely things we'd automate normally so I should go ahead and merge them as long as the tests are green.

